PR #8, opened 2022-08-12
`run_until()` spawns on its own rt

mcches:
This allows clients and test code to spawn tasks, and aligns with
the host execution model.


carllerche (approved):
It is annoying that `until` is not `'static.` If it were, we could get rid of `tokio-test`. There probably are other ways of getting rid of `tokio-test`, but this is fine.

PR #10, opened 2022-08-13
remove static requirement

carllerche:
Trying to remove the `'static` requirement while still having the `run_until` task have its own runtime.


carllerche:
I had to bring back `tokio-test` unfortunately.

mcches (approved):

PR #13, opened 2022-08-16
Clean up `Sim` reference mutability

mcches:
Allows for immutable access to `Sim` from within `run_until`.


mcches:
My first pass on this diff'd the rts and the hosts to compute clients, but I felt it would be cleaner to clearly distinguish clients from simulated hosts.

mcches:
Going in a different direction: https://github.com/tokio-rs/turmoil/pull/14

PR #14, opened 2022-08-16
Client hosts get a dedicated rt

mcches:
`Sim::run()` semantics change to run until all registered clients
finish.

### DIFF
--- a/src/software.rs
+++ b/src/software.rs
@@ -1,0 +1,40 @@
+use std::time::Duration;
+
+use tokio::{task::JoinHandle, time::Instant};
+
+use crate::rt::Rt;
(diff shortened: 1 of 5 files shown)


src/software.rs
```diff
@@ -0,0 +1,40 @@
+use std::time::Duration;
```
mcches, line 1: I'm not loving the names... any ideas?

carllerche: I would probably go w/ `Role` or something.

mcches: Updated to `Role` in the latest commit.

carllerche (approved):

PR #20, opened 2022-09-13
Move io to free functions

mcches:
This approach has a few advantages:

- Less restrictive typing: send and recv can have different
message types.
- Simpler registration and io access for hosts.
- Opens the door for free functions that allow connecting and
accepting connections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ turmoil = "0.2"
 Next, create a test file and add a test:
 
 ```rust
-use turmoil::{Builder, Io};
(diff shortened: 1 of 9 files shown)


LucioFranco (approved):

carllerche (approved):
Looks good to me. Did we ever setup CI? We should if we didn't. Also, we should include rules to check rustfmt, etc...

mcches:
> Did we ever setup CI?

Not yet. Do you have an example I can follow for how tokio repos do it?

Issue #12, opened 2022-08-13
Support bouncing a host


mcches:
**Is your feature request related to a problem? Please describe.**
No. This is new functionality.

**Describe the solution you'd like**
Hosts in the simulation are simply futures. During `run_until()` I'd like to have the ability to "bounce" a host (cancel, join and restart).

carllerche:
Do you have thoughts on how to model it? Some things should persist, e.g., a simulated "disk".

mcches:
A simulated "disk" could be part of the lib, similar `Io`. It is certainly more convenient to say a hash map is "durable" across restarts, but we could also take the stance you need to write to disk, which makes the simulation more realistic.

marcbowes:
I think we can support bounce without state for now - seems like an orthogonal feature to me. The most immediate issue is that we need to restart the host code.

PR #24, opened 2022-09-19
Bounce, impls #12

marcbowes:
You can’t bounce the bounce if you can’t even pronounce the bounce.

### DIFF
--- a/src/role.rs
+++ b/src/role.rs
@@ -1,8 +1,14 @@
-use std::time::Duration;
+use std::{pin::Pin, time::Duration};
 
+use futures::Future;
 use tokio::{task::JoinHandle, time::Instant};
 
 use crate::rt::Rt;
+
+// To support re-creation, we need to store a factory of the future that
+// represents the software. This is somewhat annoying in that it requires
+// boxxing to avoid generics.
+type Software = Box<dyn Fn() -> Pin<Box<dyn Future<Output = ()> + 'static>>>;
 
 /// Differentiates runtime fields for different host types
 pub(crate) enum Role {
@@ -10,7 +16,7 @@ pub(crate) enum Role {
     Client { rt: Rt, handle: JoinHandle<()> },
 
     /// A simulated host
-    Simulated(Rt),
+    Simulated { rt: Rt, software: Software },
 }
 
 impl Role {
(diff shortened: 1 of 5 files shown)


src/role.rs
```diff
@@ -10,22 +11,35 @@ pub(crate) enum Role {
     Client { rt: Rt, handle: JoinHandle<()> },
 
     /// A simulated host
-    Simulated(Rt),
+    Simulated {
+        rt: Rt,
+        software: Box<dyn Fn() -> Pin<Box<dyn Future<Output = ()> + 'static>>>,
```
mcches, line 16: How about some crate types for Software/Factory?

src/sim.rs
```diff
@@ -48,15 +48,24 @@ impl Sim {
...
 
     /// Register a host with the simulation
-    pub fn host<F>(&mut self, addr: impl ToSocketAddr, host: F)
+    ///
+    /// This method takes a `Fn` that builds a future, as opposed to
+    /// [`Sim::client`] which just takes a future. The reason for this is we
+    /// might restart the host, and so need to be able to call the future
+    /// multiple times.
+    ///
+    // TODO: Removing the `'static` bound on `F` would make writing some tests
```
mcches, line 63: I'm not seeing how we would do this due to constraints on `spawn_local`. Did you see a path?

src/rt.rs
```diff
@@ -53,4 +55,14 @@ impl Rt {
                 .await
         })
     }
+
+    pub(crate) fn cancel_tasks(&mut self) {
+        _ = mem::replace(&mut self.local, new_local());
```
mcches, line 60: This is a nice idea. I was messing around with abort + await at some point and it's annoying/unrealistic to crash like that.

src/role.rs
marcbowes: I'll defer this until I find myself repeating the signature. Otherwise it's just indirection for no benefit. The purpose is already named (by the field).

src/sim.rs
marcbowes: I'm not seeing why `spawn_local` is a problem!

The main issue is that the factory fn needs to live somewhere. I've moved it into the new fn that lives on the heap. That works because it has no non-static refs.

I don't know what else could possibly work.

mcches (approved):
Lgtm

src/role.rs
marcbowes: I added it as there actually *was* repetition due to type inference not working.

Issue #11, opened 2022-08-13
Support holding messages after send


mcches:
**Is your feature request related to a problem? Please describe.**
No. This is new functionality.

**Describe the solution you'd like**
During the simulation I'd like to place a "hold" on the link between two hosts. Any messages sent will remain in the queue while the "hold" is active. At a later time I'd like to remove the "hold", which allows delivery for the queued messages.

This is useful for tests that need to control the ordering of events across multiple hosts.

carllerche:
Seems like a good feature :+1:

marcbowes:
Do you think it's better to hold the messages on the link or in the host? The former seems closer to reality, but requires duplication of the inbox type. I'm inclined towards buffering on the link.

mcches:
I was considering adding a "hold" token for any messages added to the queue. The delivery logic can inspect both delivery time and if the "hold" is present.

PR #25, opened 2022-09-19
Implement hold and release, impls #11

marcbowes:
This commit introduces `turmoil::hold` and `::release` that temporarily hold packets between a sender and receiver. Unlike the partition and repair methods, held packets are not dropped. Instead, they are delivered after release is called.

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -6,8 +6,8 @@ pub(crate) struct Envelope {
     /// Who sent the message
     pub(crate) src: version::Dot,
 
-    /// When to deliver the message
-    pub(crate) deliver_at: Instant,
+    /// When to deliver the message. If `None`, the message is stuck without an ETA.
+    pub(crate) deliver_at: Option<Instant>,
 
     /// Message value
     pub(crate) message: Box<dyn Message>,
--- a/src/world.rs
+++ b/src/world.rs
@@ -1,3 +1,4 @@
+use crate::top::Embark;
 use crate::{config, Dns, Envelope, Host, Log, Message, ToSocketAddr, Topology};
 
 use indexmap::IndexMap;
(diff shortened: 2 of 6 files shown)


src/world.rs
```diff
@@ -73,6 +74,18 @@ impl World {
         self.dns.lookup(host)
     }
 
+    pub(crate) fn hold(&mut self, a: SocketAddr, b: SocketAddr) {
+        self.topology.hold(a, b);
+    }
+
+    // TODO: Should all held packets be immediately released, or should they be
```
mcches, line 81: I'll speculate and say that I don't think they need to flow through all the existing turmoil machinery. We've got two paradigms here: run some longer tests with seeded random on and run shorter tests with it off and full control (ie hold a link).

I've been fighting this a bit with streams... obviously things like message re-order and drop are impossible in a TCP. We really need to do things at a higher level for that to work, ie re-order based on distinct connections.

I could see a case here though where you "hold", then a network partition happens, then "release" and...?

mcches:
This looks good. I think we need  to make a decision on if we have turmoil modes as we introduce this and distinct connections.

mcches (approved):

src/envelope.rs
```diff
@@ -6,8 +6,8 @@ pub(crate) struct Envelope {
     /// Who sent the message
     pub(crate) src: version::Dot,
 
-    /// When to deliver the message
-    pub(crate) deliver_at: Instant,
+    /// When to deliver the message. If `None`, the message is stuck without an ETA.
+    pub(crate) deliver_at: Option<Instant>,
```
carllerche, line 10: This is fine for this PR. I would probably lean towards a custom enum for clarity whenever this code is touched again.

marcbowes: I agree, and will do that today. @mcches recommended I merge this asap to avoid rebases.

marcbowes: this was fixed via https://github.com/tokio-rs/turmoil/pull/27

PR #31, opened 2022-09-22
Fix error in README.md

marcbowes:
In a recent commit, I changed the API for starting a host to require a future factory. The example in the README broke, but the PR was still merged.

In this commit, we use Github actions to run `cargo test --verbose` PRs to main (and adhoc pushes). For now, this just tests on ubuntu as we have no platform specific code.

In addition, we run the README as a doctest. This slightly harms the README because the example is no longer copy-paste ready. In particular, the `#[test] fn simulation() {  }` chrome is now removed so as to work within the doctest framework. I think that's OK, especially since we ship real examples in test/.

Finally, the actual bug was fixed by adding a tunnel, down which no variables slide.

---

Note that I pushed this to my fork and the CI worked: https://github.com/marcbowes/turmoil/actions/runs/3102304893/jobs/5024489886

### DIFF
--- a/src/readme.rs
+++ b/src/readme.rs
@@ -1,0 +1,8 @@
+///! The only purpose of this file is to run the README.md as if it was a
+///! doctest. The reason for a standalone file is so that any failures show up
+///! clearly as 'readme' failures, as opposed to tests in lib.rs or some other
+///! file.
+///!
+///! https://blog.guillaume-gomez.fr/articles/2020-03-07+cfg%28doctest%29+is+stable+and+you+should+use+it
+
+doc_comment::doctest!("../README.md");
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,33 @@
+name: CI Build
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    name: Build and Test
+    runs-on: ${{ matrix.os }}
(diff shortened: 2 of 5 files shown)


src/readme.rs
```diff
@@ -0,0 +1,8 @@
+///! The only purpose of this file is to run the README.md as if it was a
+///! doctest. The reason for a standalone file is so that any failures show up
+///! clearly as 'readme' failures, as opposed to tests in lib.rs or some other
+///! file.
+///!
+///! https://blog.guillaume-gomez.fr/articles/2020-03-07+cfg%28doctest%29+is+stable+and+you+should+use+it
+
+doc_comment::doctest!("../README.md");
```
LucioFranco, line 8: I feel like this could be a `tests/readme.rs` and not have to live in the src dir?

LucioFranco (approved):
LGTM, just one question

mcches:
Can we also add a build badge to the README?

e.g. https://github.com/tokio-rs/tokio/blob/master/README.md?plain=1#L17

mcches (approved):

src/readme.rs
marcbowes: Would be nice, but that didn't work for me. I assume that's because by default, `cargo test` does not do doctest in the tests/ directory? Maybe there is a flag to enable that, but even so I like that it just works right now.

marcbowes:
> Can we also add a build badge to the README?
> 
> e.g. https://github.com/tokio-rs/tokio/blob/master/README.md?plain=1#L17

I think we can do that under the related issue :)

PR #34, opened 2022-09-29
Some cleanup for existing networking

mcches:
- s/send/embark to better self document what happens
- `send` bumps host version: Originally only recv bumps this, but
connections will bump on connect and accept, so this prepares for
that.
- minor refactoring for send and recv paths
- s/Arc/Rc

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -4,7 +4,7 @@ use crate::{version, Envelope, Message};
 use indexmap::IndexMap;
 use std::collections::VecDeque;
 use std::net::SocketAddr;
-use std::sync::Arc;
+use std::rc::Rc;
 use tokio::sync::Notify;
 use tokio::time::{Duration, Instant};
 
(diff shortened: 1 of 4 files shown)


src/host.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) struct Host {
 }
 
 impl Host {
-    pub(crate) fn new(addr: SocketAddr, now: Instant, notify: Arc<Notify>) -> Host {
+    pub(crate) fn new(addr: SocketAddr, now: Instant, notify: Rc<Notify>) -> Host {
```
marcbowes, line 33: Nice

src/host.rs
```diff
@@ -54,7 +54,7 @@ impl Host {
         }
     }
 
-    pub(crate) fn send(
+    pub(crate) fn embark(
```
marcbowes, line 57: I appreciate the name change, and this is better than `send`, but it still sucks.

marcbowes (approved):

src/host.rs
mcches: Name or structure?

PR #35, opened 2022-10-03
Add support for simulated TCP connections

mcches:
The `net` module mirrors [tokio::net](https://docs.rs/tokio/latest/tokio/net/index.html).

Hosts now have the ability to bind their current `SocketAddr`,
which sets up a listener queue for accepting connections.

Connections are identified using a `SocketPair`, which is
a tuple of `version::Dot`. Note that each network operation
bumps the host's version number ensuring this is unique.

Support for "bounce" and "drop" (via RST) will follow in a subsequent
change set.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -1,14 +1,23 @@
 use crate::envelope::DeliveryInstructions;
+use crate::net::{Segment, SocketPair, StreamEnvelope, Syn};
 use crate::{version, Envelope, Message};
 
 use indexmap::IndexMap;
 use std::collections::VecDeque;
+use std::io;
 use std::net::SocketAddr;
 use std::rc::Rc;
+use std::sync::Arc;
 use tokio::sync::Notify;
 use tokio::time::{Duration, Instant};
 
-/// A host in the simulated network
+/// A host in the simulated network.
+///
+/// Hosts support two networking modes:
+/// - Datagram via [`crate::io::send`] and [`crate::io::recv`]
+/// - Stream via [`crate::net::Listener`] and [`crate::net::Stream`]
+///
+/// Both modes may be used by host software simulatanesouly.
 pub(crate) struct Host {
     /// Host address
     pub(crate) addr: SocketAddr,
@@ -17,16 +26,34 @@ pub(crate) struct Host {
     /// network".
     inbox: IndexMap<SocketAddr, VecDeque<Envelope>>,
 
-    /// Signaled when a message becomes available to receive
+    /// Signaled when a message becomes available to receive.
     pub(crate) notify: Rc<Notify>,
 
-    /// Current instant at the host
+    /// Optional accept queue; set if the host is bound. This simulates a server
+    /// socket.
+    listener: Option<Inbox<Envelope>>,
+
+    /// In-flight messages for active connections. Some of these may still be
+    /// "on the network".
+    connections: IndexMap<SocketPair, Inbox<StreamEnvelope>>,
+
+    /// Current instant at the host.
     pub(crate) now: Instant,
 
     epoch: Instant,
 
-    /// Current host version. This is incremented each time a message is received.
+    /// Current host version. This is incremented each time a network operation
+    /// occurs.
     pub(crate) version: u64,
+}
+
+/// A simple unbounded channel.
+struct Inbox<T> {
+    /// Pending connections
+    deque: VecDeque<T>,
+
+    /// Signaled when an item is available
+    notify: Arc<Notify>,
 }
 
 impl Host {
@@ -35,15 +62,53 @@ impl Host {
             addr,
             inbox: IndexMap::new(),
             notify,
+            listener: None,
+            connections: IndexMap::new(),
             now,
             epoch: now,
             version: 0,
         }
     }
 
+    /// Creates a new listener queue, which is bound to the host's `addr`.
+    ///
+    /// This is called by `Listener::bind()` and the returned `Notify` is used
+    /// to signal when connections are available to accept.
+    // TODO: Support binding to multiple ports
+    pub(crate) fn bind(&mut self) -> io::Result<Arc<Notify>> {
+        if self.listener.is_some() {
+            return Err(io::Error::new(
+                io::ErrorKind::AddrInUse,
+                self.addr.to_string(),
+            ));
+        }
+
+        let notify = Arc::new(Notify::new());
+
+        self.listener.replace(Inbox {
+            deque: VecDeque::new(),
+            notify: notify.clone(),
+        });
+
+        Ok(notify)
+    }
+
+    /// Unbind the host, dropping all pending connections.
+    pub(crate) fn unbind(&mut self) {
+        self.listener.take();
+    }
+
     /// Returns how long the host has been executing for in virtual time
     pub(crate) fn elapsed(&self) -> Duration {
         self.now - self.epoch
+    }
+
+    /// Bump the version for this host and return a dot.
+    ///
+    /// Called when a host establishes a new connection with a remote peer.
+    pub(crate) fn bump(&mut self) -> version::Dot {
+        self.version += 1;
+        self.dot()
     }
 
     /// Returns a dot for the host at its current version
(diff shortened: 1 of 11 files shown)


src/host.rs
```diff
@@ -1,14 +1,23 @@
...
 use tokio::time::{Duration, Instant};
 
-/// A host in the simulated network
+/// A host in the simulated network.
+///
+/// Hosts support two networking modes:
+/// - Datagram via [`crate::io::send`] and [`crate::io::recv`]
+/// - Stream via [`crate::net::Listener`] and [`crate::net::Stream`]
+///
+/// Both modes may be used by host software simulatanesouly.
```
mcches, line 20: Lots of duplication here, but it's all internal. I'd like to get the public part and tests in and we can clean this up.

src/net/stream.rs
```diff
@@ -0,0 +1,157 @@
...
+
+                let notify = Arc::clone(&self.notify);
+                self.read_fut
+                    .get_or_insert(ReusableBoxFuture::new(
+                        async move { notify.notified().await },
+                    ))
+            }
+        };
+
+        match read_fut.poll(cx) {
```
LucioFranco, line 79: not sure if you know but there is a https://doc.rust-lang.org/std/task/macro.ready.html macro that might make this code easier. Its kinda the same as `?` for Result.

src/net/stream.rs
```diff
@@ -0,0 +1,157 @@
...
+            world.embark_on(self.pair, Segment::Data(bytes))
+        });
+
+        Poll::Ready(Ok(buf.len()))
+    }
+}
+
+impl Drop for Stream {
+    fn drop(&mut self) {}
+}
```
LucioFranco, line 129: I assume this is here to add Rst? Might wanna add a TODO?

LucioFranco (approved):
LGTM overall left some nits etc

src/host.rs
```diff
@@ -54,6 +121,67 @@ impl Host {
...
+            listener.deque.push_back(Envelope {
+                src,
+                instructions,
+                message: Box::new(syn),
+            });
+
+            listener.notify.notify_one();
+        }
+
+        // notify drops, connection is refused
```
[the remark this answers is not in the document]
mcches: The notify within the `Syn` drops. The client is awaiting that, so by dropping we get connection refused naturally. I'll add a bit more to this.

src/host.rs
```diff
@@ -54,6 +121,67 @@ impl Host {
...
+        let inbox = self.connections.entry(pair).or_insert_with(|| Inbox {
+            deque: VecDeque::new(),
+            notify: Arc::new(Notify::new()),
+        });
+
+        inbox.notify.clone()
+    }
+
+    pub(crate) fn register_connection(&mut self, pair: SocketPair) {
+        assert!(self
```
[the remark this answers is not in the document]
mcches: Good call.

src/net/mod.rs
```diff
@@ -0,0 +1,75 @@
+//! This module contains the simulated TCP/UDP networking types.
+//!
+//! They mirror [tokio::net](https://docs.rs/tokio/latest/tokio/net/) to provide
+//! a high fidelity implementation.
```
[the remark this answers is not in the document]
mcches: I can't tell the difference from this diff?

src/net/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub(crate) instructions: DeliveryInstructions,
+
+    /// Segment type
+    pub(crate) segment: Segment,
+}
+
+/// Message types for established connections.
+#[derive(Debug)]
+pub(crate) enum Segment {
+    Data(Bytes),
```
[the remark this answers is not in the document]
mcches: Yup, and Fin.

src/net/stream.rs
mcches: I looked at the one in futures, but let me revisit as I agree avoiding some nesting would be nice in here.

src/net/stream.rs
mcches: I'll just delete this for now.

src/net/listener.rs
```diff
@@ -0,0 +1,59 @@
...
+use tokio::sync::Notify;
+
+use crate::world::World;
+
+use super::Stream;
+
+/// A simulated socket server, listening for connections.
+///
+/// All methods must be called from a host within a Turmoil simulation.
+pub struct Listener {
```
carllerche, line 12: Do we want to call this `TcpListener`? I know it doesn't *actually* use TCP, but it is meant to follow Tokio's Tcp* type semantics.

carllerche (approved):
LGTM, I left a couple of thoughts inline (not blockers).

src/net/listener.rs
mcches: I'm open to either. What's missing in turmoil is guidance or an actual abstraction to bind to either implementation. I'm taking the stance that it's up to the consuming application to define that, because they might have specific requirements on things like how to bind.

@marcbowes and I briefly discussed adding an example that contains an example client/server with a way to choose either turmoil or tokio::net.

src/net/mod.rs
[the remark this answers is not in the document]
mcches: Oh, 🤦

src/host.rs
marcbowes, line 20: simultaneously

src/host.rs
```diff
@@ -54,6 +125,72 @@ impl Host {
...
+                self.bump_version();
+                ret
+            }
+            _ => None,
+        }
+    }
+
+    // If the host is not bound we simply do nothing, dropping the `Syn`. The
+    // peer who initiated the connection is awaiting the receiver of the syn's
+    // oneshot, which triggers a "connection refused" error.
```
marcbowes, line 147: 👍

src/host.rs
```diff
@@ -54,6 +125,72 @@ impl Host {
...
+                message: Box::new(syn),
+            });
+
+            listener.notify.notify_one();
+        }
+    }
+
+    /// Finalize the connection, returning a `Notify` to build the stream.
+    ///
+    /// This is idempotent. If the connection has already been initialized, the
```
marcbowes, line 167: Why does it need to be idempotent?

src/host.rs
```diff
@@ -54,6 +125,72 @@ impl Host {
...
+            deque: VecDeque::new(),
+            notify: Arc::new(Notify::new()),
+        });
+
+        inbox.notify.clone()
+    }
+
+    pub(crate) fn register_connection(&mut self, pair: SocketPair) {
+        assert!(
+            self.connections
```
marcbowes, line 180: I don't think this reads very well.

src/host.rs
```diff
@@ -70,11 +198,31 @@ impl Host {
             instructions,
             message,
         });
-        self.version += 1;
 
         self.notify.notify_one();
     }
 
+    pub(crate) fn embark_on(
```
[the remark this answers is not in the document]
marcbowes: Embark, like getting on a ship. You're not yet at your destination - you're just on the ship. Maybe I picked this name because I was thinking along the line of ports.. Either way, I don't love the name, but can't think of anything better. Maybe enqueue?

Send was ambiguous in so many ways. Was the host sending something (no.. it was preparing to receive something). When the method returns is it (the message) then sent (no.. it's in a pipe somewhere).

src/host.rs
```diff
@@ -54,6 +125,72 @@ impl Host {
         }
     }
 
+    pub(crate) fn accept(&mut self) -> Option<Envelope> {
+        let now = Instant::now();
+        let deque = &mut self.listener.as_mut()?.deque;
+
+        match deque.front() {
+            Some(Envelope {
+                instructions: DeliveryInstructions::DeliverAt(time),
```
marcbowes, line 134: There are three hosts, A, B & C. A can't reach B, but tries to connect. C also tries to connect to B. What happens?

I think this is broken, because syn(A:B) is at the front of the deque but is not eligible for delivery. If so, please open an issue. When we fix it, we should add a regression test.

src/host.rs
```diff
@@ -54,6 +125,72 @@ impl Host {
...
+            };
+
+            listener.deque.push_back(Envelope {
+                src,
+                instructions,
+                message: Box::new(syn),
+            });
+
+            listener.notify.notify_one();
+        }
```
marcbowes, line 162: else .. what? Everything is fine, or this represents a bug?

src/host.rs
```diff
@@ -97,29 +255,81 @@ impl Host {
...
+
+        let deque = self
+            .connections
+            .get_mut(&pair)
+            .map(|i| &mut i.deque)
+            .expect("no connection");
+
+        match deque.front() {
+            Some(StreamEnvelope {
+                instructions: DeliveryInstructions::DeliverAt(time),
```
marcbowes, line 286: Unlike the other comment (for accepting connections), I think this code is correct. Just flagging it in case I'm wrong.

src/host.rs
```diff
@@ -97,29 +255,81 @@ impl Host {
...
+    }
+
     /// Releases all messages previously received from [`peer`]. These messages
     /// may be received immediately (on the next call to `[Host::recv]`).
     pub(crate) fn release(&mut self, peer: SocketAddr) {
         let now = Instant::now();
-        let deque = self.inbox.entry(peer).or_default();
 
-        for envelope in deque {
+        if let Some(listener) = self.listener.as_mut() {
```
marcbowes, line 302: I don't love that we have to release from two places. I'm not worried about the small amount of code duplication, but I do think it's a smell that the abstractions aren't quite right. All internal though..

src/host.rs
```diff
@@ -135,6 +345,30 @@ impl Host {
     pub(crate) fn tick(&mut self, now: Instant) {
         self.now = now;
 
+        if let Some(listener) = self.listener.as_ref() {
```
marcbowes, line 348: Ah, more duplication! See previous comment.

src/net/stream.rs
```diff
@@ -0,0 +1,144 @@
...
+        let notify = World::current(|world| world.current_host_mut().finish_connect(pair));
+
+        Ok(Self::new(pair, notify))
+    }
+
+    fn poll_read_priv(
+        &mut self,
+        cx: &mut Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<io::Result<()>> {
```
marcbowes, line 61: I have not reviewed this carefully. Just noting that.

src/host.rs
mcches: This is leftover from an issue I had where we send right after accepting. Once ack'd the inbox is setup, so `finish_connect()` from the initiator simply needs the `Notify` to build the stream. I'll clean this up.

src/host.rs
mcches: Agree, how does this look?

```rust
pub(crate) fn register_connection(&mut self, pair: SocketPair) {
    let inbox = Inbox {
        deque: VecDeque::new(),
        notify: Arc::new(Notify::new()),
    };

    let contains_pair = self.connections.insert(pair, inbox).is_some();

    assert!(!contains_pair, "{:?} is already registered", pair);
}
```

Note that this may change when I clean up `finish_connect()`.

src/host.rs
mcches: Everything is fine. We drop the `Syn`, which holds the sender and the caller gets connection refused awaiting the receiver. It's described above. I had a comment on line 163, but it caused confusion for earlier reviewers.

src/host.rs
mcches: Yup, we've got ordering guarantees within the connection.

src/host.rs
mcches: That sounds broken to me, but I'll write the test and report back to verify. It shouldn't be too tough to fix if it is.

src/host.rs
mcches: Agree.

src/host.rs
mcches: Yeah... embark, recv, release and tick are all pretty much copies. It should all be confined within host and we can restructure easily.

Issue #36, opened 2022-10-03
Support client and host software errors


mcches:
Currently, we only have panic to trigger failure during simulation runs. This makes writing both the test and host software a little clunky, as we can't `?` return on err.

To make the experience better, we can define a dynamic Error type and have both client and hosts supply a future that aligns. On each `run()` iteration, we can check if any host finishes with an error, end the simulation and return the error.

e.g.
```rust
pub type TurmoilResult<T = ()> = std::result::Result<T, Box<dyn std::error::Error>>;
```

PR #37, opened 2022-10-04
Add a turmoil `Result` type

mcches:
This faciliates writing fallible tests via clients without forcing
panics to signal failure.

It allows early return for any `Error` type via `?`.


mcches:
Doing this for host is a bit trickier due to crash, bounce and subsequent run invocations.

mcches:
Issue #36

src/error.rs
```diff
@@ -0,0 +1,8 @@
+/// A specialized [`Result`] type for turmoil simulations.
+///
+/// This type is generally useful for fallible test cases, i.e. where you want
+/// to use the `?` operator to fail the test rather than writing unwrap
+/// everywhere.
+///
+/// [`Result`]: std::result::Result
+pub type Result<T = ()> = std::result::Result<T, Box<dyn std::error::Error>>;
```
carllerche, line 8: You could consider depending on [`anyhow`](https://docs.rs/anyhow/latest/anyhow/). This is also fine though.

carllerche (approved):

src/error.rs
mcches: Thanks for the suggestion. I'm going to stick with our own since it is a public API.

marcbowes: I agree with our own type. However, I'd prefer `BoxResult` as the name.

marcbowes (commented):
Nice

PR #40, opened 2022-10-05
Follow up on late comments to #35

mcches:
Most notably, this fixes a bug in the accept queue where syn's not
eligible for delivery block other connections from proceeding.

A regression test is added for this specific scenario.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -14,10 +14,10 @@ use tokio::time::{Duration, Instant};
 /// A host in the simulated network.
 ///
 /// Hosts support two networking modes:
-/// - Datagram via [`crate::io::send`] and [`crate::io::recv`]
-/// - Stream via [`crate::net::Listener`] and [`crate::net::Stream`]
+/// - Datagram
+/// - Stream
 ///
-/// Both modes may be used by host software simulatanesouly.
+/// Both modes may be used by host software simultaneously.
 pub(crate) struct Host {
     /// Host address
     pub(crate) addr: SocketAddr,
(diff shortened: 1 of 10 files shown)


src/host.rs
```diff
@@ -127,19 +127,25 @@ impl Host {
...
+        // avoid front-of-line blocking.
+        for (index, envelope) in deque.iter().enumerate() {
+            match envelope {
+                Envelope {
+                    instructions: DeliveryInstructions::DeliverAt(time),
+                    ..
+                } if *time <= now => {
+                    self.bump_version();
+                    let deque_mut = &mut self.listener.as_mut()?.deque;
+                    return deque_mut.remove(index);
```
marcbowes, line 142: It's wild that this compiles. At least two things are happening:

1. The borrow on L130 is being released, so as to allow a mut ref.
2. And this is safe because L142 returns and so the immu borrow is never taken again.

Really awesome stuff.

tests/net.rs
```diff
@@ -154,6 +154,44 @@ fn hold_and_release_once_connected() -> turmoil::Result {
...
+                .await
+                .is_err()
+        );
+
+        Ok(())
+    });
+
+    // C:B should succeed, and should not be blocked behind the A:B, which is
+    // not eligible for delivery
+    sim.client("C", async {
```
marcbowes, line 186: Did this test fail before?

I'm not entirely sure how the scheduling works out here. What guarantees the hold (L173) happens before C:B syn on L187?

src/host.rs
mcches: Right!? I too stared at this in disbelief.

tests/net.rs
mcches: It did fail due to deterministic host registration and execution. I'll add a waiter to ensure the order in case we change something.

marcbowes (approved):

PR #42, opened 2022-10-07
Add drop and shutdown for TcpStream

mcches:


src/net/stream.rs
```diff
@@ -87,31 +98,72 @@ impl TcpStream {
...
         match World::current(|world| world.recv_on(pair)) {
-            Some(seg) => match seg {
-                Segment::Data(bytes) => {
-                    buf.put_slice(bytes.as_ref());
-                    return Some(Ok(()));
-                }
-            },
+            Some(seg) => {
+                let bytes = match seg {
+                    Segment::Data(bytes) => bytes,
```
LucioFranco, line 114: What happens if a data frame of zero bytes will that trigger some other eof behavior?

LucioFranco (approved):

src/net/stream.rs
mcches: Good catch, that is broken. We should no-op the write side to make this impossible.

mcches: See [d8a156b](https://github.com/tokio-rs/turmoil/pull/42/commits/d8a156b34d26c7e91ce5b9459a51afc8a3b42354) for fix and test.

LucioFranco (approved):

mcches:
Rolling this into a bigger refactor.

PR #48, opened 2022-10-25
Simulated TCP streams

mcches:
`TcpListener` and `TcpStream` are implemented with the minimal APIs mirrored
from `tokio::net` to be functional.

Tracing is updated to use structured k/v pairs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,37 +32,19 @@ turmoil = "0.2"
 Next, create a test file and add a test:
 
 ```rust
-use std::net::{IpAddr, Ipv4Addr};
-
-use turmoil::{lookup, net, Builder};
-
-let port = 1234;
-let unspecified = IpAddr::from(Ipv4Addr::UNSPECIFIED);
-
-let mut sim = Builder::new().build();
+let mut sim = turmoil::Builder::new().build();
 
 // register a host
 sim.host("server", || async move {
-    let sock = net::UdpSocket::bind((unspecified, port)).await.unwrap();
-
-    loop {
-        let mut buf = vec![0; 12];
-        let (_, origin) = sock.recv_from(&mut buf).await.unwrap();
-
-        let _ = sock.send_to(&buf, origin).await;
-    }
+    // host software goes here
 });
 
-// register a client (this is the test code)
+// register a client
 sim.client("client", async move {
-    let sock = net::UdpSocket::bind((unspecified, port)).await?;
+    // dns lookup for "server"
+    let addr = turmoil::lookup("server");
 
-    let server_addr = lookup("server");
-    sock.send_to(b"hello, world", (server_addr, port)).await?;
-
-    let mut buf = vec![0; 12];
-    let _ = sock.recv_from(&mut buf).await?;
-    assert_eq!(b"hello, world", &buf[..]);
+    // test code goes here
 
     Ok(())
 });
@@ -71,7 +53,7 @@ sim.client("client", async move {
 sim.run();
 ```
 
-For more examples, check out the [tests](tests) directory.
+For more examples (with simulated networking), check out the [tests](tests) directory.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,9 @@ sim.client("client", async move {
 sim.run();
 ```
 
-For more examples (with simulated networking), check out the [tests](tests) directory.
+See `ping_pong` in [udp.rs](tests/udp.rs) for a TCP example.
+
+For more examples, check out the [tests](tests) directory.
 
 ## License
 
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -61,6 +61,7 @@ impl ToIpAddr for IpAddr {
     }
 }
 
+// Hostname and port
 impl ToSocketAddr for (String, u16) {
     fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
         (&self.0[..], self.1).to_socket_addr(dns)
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -11,7 +11,7 @@ pub trait ToIpAddr {
 }
 
 pub trait ToSocketAddr {
-    fn to_socket_addr(&self) -> SocketAddr;
+    fn to_socket_addr(&self, dns: &Dns) -> SocketAddr;
 }
 
 impl Dns {
@@ -61,14 +61,29 @@ impl ToIpAddr for IpAddr {
     }
 }
 
+impl ToSocketAddr for (String, u16) {
+    fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
+        (&self.0[..], self.1).to_socket_addr(dns)
+    }
+}
+
+impl<'a> ToSocketAddr for (&'a str, u16) {
+    fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
+        match dns.names.get(self.0) {
+            Some(ip) => (*ip, self.1).into(),
+            None => panic!("no hostname found for ip address"),
+        }
+    }
+}
+
 impl ToSocketAddr for SocketAddr {
-    fn to_socket_addr(&self) -> SocketAddr {
+    fn to_socket_addr(&self, _: &Dns) -> SocketAddr {
         *self
     }
 }
 
 impl ToSocketAddr for (IpAddr, u16) {
-    fn to_socket_addr(&self) -> SocketAddr {
+    fn to_socket_addr(&self, _: &Dns) -> SocketAddr {
         (*self).into()
     }
 }
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -1,7 +1,7 @@
 use std::{fmt::Display, net::SocketAddr};
 
 use bytes::Bytes;
-use futures::channel::oneshot;
+use tokio::sync::oneshot;
 
 #[derive(Debug)]
 pub(crate) struct Envelope {
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -23,7 +23,7 @@ pub(crate) struct Datagram(pub Bytes);
 ///
 /// We implement just enough to ensure fidelity and provide knobs for real world
 /// scenarios, but we skip a ton of complexity (e.g. checksums, flow control,
-/// etc) because said complexity isn't useful needed in tests.
+/// etc) because said complexity isn't useful in tests.
 #[derive(Debug)]
 pub(crate) enum Segment {
     Syn(Syn),
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -19,25 +19,17 @@ pub(crate) enum Protocol {
 #[derive(Debug)]
 pub(crate) struct Datagram(pub Bytes);
 
-/// This is a parody of TCP.
+/// This is a simplification of real TCP.
 ///
-/// We implement just enough to make it realistic, but we skip a ton of
-/// complexity (e.g. checksums, flow control, etc).
+/// We implement just enough to ensure fidelity and provide knobs for real world
+/// scenarios, but we skip a ton of complexity (e.g. checksums, flow control,
+/// etc) because said complexity isn't useful needed in tests.
 #[derive(Debug)]
 pub(crate) enum Segment {
     Syn(Syn),
-    Data(u64, StreamData),
+    Data(u64, Bytes),
     Fin(u64),
     Rst,
-}
-
-#[derive(Debug)]
-pub(crate) struct StreamData(pub Bytes);
-
-impl StreamData {
-    pub(crate) fn eof() -> Self {
-        Self(Bytes::new())
-    }
 }
 
 #[derive(Debug)]
@@ -56,8 +48,7 @@ impl Display for Protocol {
 
 impl Display for Datagram {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "UDP ")?;
-        hex(&self.0, f)
+        hex("UDP", &self.0, f)
     }
 }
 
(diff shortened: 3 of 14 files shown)


src/envelope.rs
```diff
@@ -1,6 +1,7 @@
 use std::{fmt::Display, net::SocketAddr};
 
 use bytes::Bytes;
+use futures::channel::oneshot;
```
davidbarsky, line 4: can [Tokio's oneshot channel](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html) be used instead instead? it doesn't depend on the tokio runtime being active, so if you're worried about circular dependencies, it's not an issue here.

src/host.rs
```diff
@@ -68,13 +106,19 @@ impl Udp {
...
+    }
+
     pub(crate) fn bind(&mut self, addr: SocketAddr) -> io::Result<UdpSocket> {
         let (tx, rx) = mpsc::unbounded_channel();
 
         if self.binds.insert(addr, tx).is_some() {
             return Err(io::Error::new(io::ErrorKind::AddrInUse, addr.to_string()));
         }
 
+        trace!(?addr, protocol = %"UDP", "Bind");
```
davidbarsky, line 120: suggestion to avoid formatting a string with display, but I think this is better done as a span.
```suggestion
        trace!(?addr, protocol = "UDP", "Bind");
```

src/host.rs
```diff
@@ -88,6 +132,159 @@ impl Udp {
     }
 
     pub(crate) fn unbind(&mut self, addr: SocketAddr) {
-        assert!(self.binds.remove(&addr).is_some(), "unknown bind {}", addr)
+        assert!(self.binds.remove(&addr).is_some(), "unknown bind {}", addr);
+
+        trace!(?addr, protocol = %"UDP", "Unbind");
```
davidbarsky, line 137: same suggestion as above, but i think this is better done as a span.

src/host.rs
```diff
@@ -88,6 +132,159 @@ impl Udp {
...
+        let sock = self.sockets.get_mut(&pair)?;
+        Some(sock.assign_seq())
+    }
+
+    fn receive_from_network(
+        &mut self,
+        src: SocketAddr,
+        dst: SocketAddr,
+        segment: Segment,
+    ) -> Option<Protocol> {
```
davidbarsky, line 242: i'm not convinced that this should be `Option`—maybe a Result instead?

src/net/listener.rs
```diff
@@ -1,4 +1,88 @@
...
+
+    /// Accepts a new incoming connection from this listener.
+    ///
+    /// This function will yield once a new TCP connection is established. When
+    /// established, the corresponding [`TcpStream`] and the remote peer’s
+    /// address will be returned.
+    pub async fn accept(&self) -> Result<(TcpStream, SocketAddr)> {
+        loop {
+            let (syn, origin) = self.receiver.borrow_mut().recv().await.unwrap();
+            trace!(dst = ?origin, src = ?self.local_addr, protocol = %"TCP SYN", "Recv");
```
davidbarsky, line 57: i think this should _arguably_ be a span instead of an event

src/net/stream.rs
```diff
@@ -1,4 +1,188 @@
...
+        Self {
+            pair,
+            receiver,
+            is_closed: false,
+            is_shutdown: false,
+        }
+    }
+
+    /// Opens a TCP connection to a remote host.
+    pub async fn connect<A: ToSocketAddr>(addr: A) -> Result<TcpStream> {
```
davidbarsky, line 47: i think a span would be perfect here.

src/net/stream.rs
```diff
@@ -1,4 +1,188 @@
...
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<()>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn poll_shutdown(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<()>> {
+        self.poll_shutdown_priv()
+    }
+}
```
davidbarsky, line 177: i think adding spans to each of these impls would be quite good useful, i think. you'd probably want to have a single persistent span corresponding to `TcpStream` that gets entered on each method call.

src/net/udp.rs
```diff
@@ -78,26 +72,19 @@ impl UdpSocket {
     pub async fn recv_from(&self, buf: &mut [u8]) -> Result<(usize, SocketAddr)> {
         let (datagram, origin) = self.rx.borrow_mut().recv().await.unwrap();
 
-        // Grab the trace before we mutate things
-        let trace = format!("Recv {} {} {}", self.local_addr, origin, datagram);
+        trace!(dst = ?self.local_addr, src = ?origin, protocol = %datagram, "Recv");
```
davidbarsky, line 75: this be a candidate for a span, maybe.

src/envelope.rs
mcches: Oops, messed up the import. Tokio is what we want.

src/host.rs
```diff
@@ -88,6 +132,159 @@ impl Udp {
...
+        match segment {
+            Segment::Syn(syn) => {
+                // If bound, queue the syn; else we drop the syn triggering
+                // connection refused on the client.
+                if let Some(b) = self.binds.get_mut(&dst) {
+                    let _ = b.send((syn, src));
```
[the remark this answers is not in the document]
mcches: It's an unreachable path. Dropping the listener unbinds the port.

src/host.rs
mcches: I could see a Result with an Err(Rst) reading nicely here.

src/net/listener.rs
```diff
@@ -1,4 +1,88 @@
...
+            local_addr,
+            receiver: RefCell::new(receiver),
+        }
+    }
+
+    /// Creates a new TcpListener, which will be bound to the specified address.
+    ///
+    /// The returned listener is ready for accepting connections.
+    ///
+    /// Only 0.0.0.0 is currently supported.
```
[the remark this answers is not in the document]
mcches: I'm punting on things like loopback, and other complexity for now. I'm not sure how any IP address would work for binding. Hosts are discoverable via a static IP, assigned by the sim when created.

src/net/stream.rs
```diff
@@ -1,4 +1,188 @@
-/// A simulated connection between two hosts.
+use std::{
+    io::{self, Result},
+    pin::Pin,
+    task::{ready, Context, Poll},
+};
+
+use bytes::{Buf, Bytes};
+use futures::channel::oneshot;
```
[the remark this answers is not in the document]
mcches: Oops, messed up the import. Tokio is what we want.

mcches:
@davidbarsky RE: Spans. That makes a lot of sense. I think tracing needs a couple follow ups, but I prefer to address that outside of this change set.

src/host.rs
mcches: I was going for consistency, `protocol = UDP` when done via `Display` vs. `protocol = "UDP"` when done manually.

LucioFranco:
@davidbarsky for spans can those just be the instrument macro?

mcches:
> @davidbarsky RE: Spans. That makes a lot of sense. I think tracing needs a couple follow ups, but I prefer to address that outside of this change set.

https://github.com/tokio-rs/turmoil/issues/49

README.md
```diff
@@ -32,37 +32,19 @@ turmoil = "0.2"
...
 sim.host("server", || async move {
-    let sock = net::UdpSocket::bind((unspecified, port)).await.unwrap();
-
-    loop {
-        let mut buf = vec![0; 12];
-        let (_, origin) = sock.recv_from(&mut buf).await.unwrap();
-
-        let _ = sock.send_to(&buf, origin).await;
-    }
+    // host software goes here
```
marcbowes, line 39: +1

README.md
```diff
@@ -71,7 +53,7 @@ sim.client("client", async move {
 sim.run();
 ```
 
-For more examples, check out the [tests](tests) directory.
+For more examples (with simulated networking), check out the [tests](tests) directory.
```
marcbowes, line 56: Can you link directly to an example, such as the echo example?

src/dns.rs
```diff
@@ -11,7 +11,7 @@ pub trait ToIpAddr {
 }
 
 pub trait ToSocketAddr {
-    fn to_socket_addr(&self) -> SocketAddr;
+    fn to_socket_addr(&self, dns: &Dns) -> SocketAddr;
```
marcbowes, line 14: Is it concerning that this doesn't line up with tokio's signature (where DNS is global)?

src/dns.rs
```diff
@@ -61,14 +61,29 @@ impl ToIpAddr for IpAddr {
     }
 }
 
+impl ToSocketAddr for (String, u16) {
```
marcbowes, line 64: A comment saying we expect the u16 to be a port would be helpful, even if obvious.

src/envelope.rs
```diff
@@ -11,31 +12,82 @@ pub(crate) struct Envelope {
...
+    Tcp(Segment),
     Udp(Datagram),
 }
 
 #[derive(Debug)]
 pub(crate) struct Datagram(pub Bytes);
 
+/// This is a parody of TCP.
+///
+/// We implement just enough to make it realistic, but we skip a ton of
```
marcbowes, line 24: Realistic is probably the wrong term (& I know you copied this comment from some nonsense I wrote).

I think the goal is fidelity and the ability to simulate real world scenarios. So, checksums are not something we implement because in the real world TCP does checksums for us, and so there is nothing for us to test because we trust TCP will do its job.

src/envelope.rs
```diff
@@ -11,31 +12,82 @@ pub(crate) struct Envelope {
...
+    Fin(u64),
+    Rst,
+}
+
+#[derive(Debug)]
+pub(crate) struct StreamData(pub Bytes);
+
+impl StreamData {
+    pub(crate) fn eof() -> Self {
+        Self(Bytes::new())
```
marcbowes, line 39: I don't know how, but I know this will byte (sic) us.

src/host.rs
```diff
@@ -40,13 +49,43 @@ impl Host {
         self.now - self._epoch
     }
 
-    pub(crate) fn receive_from_network(&mut self, envelope: Envelope) {
+    pub(crate) fn assign_ephemeral_port(&mut self) -> u16 {
+        // re-load
+        if self.next_ephemeral_port == 65535 {
+            self.next_ephemeral_port = 1024;
+        }
```
marcbowes, line 56: Should this not be in the loop?

src/host.rs
```diff
@@ -88,6 +133,151 @@ impl Udp {
...
-        assert!(self.binds.remove(&addr).is_some(), "unknown bind {}", addr)
+        assert!(self.binds.remove(&addr).is_some(), "unknown bind {}", addr);
+
+        trace!(?addr, protocol = %"UDP", "Unbind");
+    }
+}
+
+pub(crate) struct Tcp {
+    /// Bound server sockets
+    binds: IndexMap<SocketAddr, mpsc::UnboundedSender<(Syn, SocketAddr)>>,
```
marcbowes, line 144: Why unbound?

I would put a limit on it, probably matching some linux standards, and note it's a TODO to make this configurable if need be.

src/host.rs
```diff
@@ -88,6 +133,151 @@ impl Udp {
...
+struct StreamSocket {
+    buf: IndexMap<u64, StreamData>,
+    next_send_seq: u64,
+    recv_seq: u64,
+    sender: mpsc::UnboundedSender<StreamData>,
+}
+
+impl StreamSocket {
+    fn new() -> (Self, mpsc::UnboundedReceiver<StreamData>) {
+        let (tx, rx) = mpsc::unbounded_channel();
```
marcbowes, line 159: Unbound question again

src/host.rs
```diff
@@ -88,6 +133,151 @@ impl Udp {
...
+            self.recv_seq += 1;
+
+            let data = self.buf.remove(&self.recv_seq).unwrap();
+            let _ = self.sender.send(data);
+        }
+    }
+}
+
+impl Tcp {
+    fn install() -> Self {
```
marcbowes, line 195: why not "new". Are you trying to signify it's a. singleton per host?

src/host.rs
```diff
@@ -88,6 +133,151 @@ impl Udp {
...
+    }
+
+    // Buffer and re-order received segments by `seq` as the network may deliver
+    // them out of order.
+    fn buffer(&mut self, seq: u64, data: StreamData) {
+        assert!(
+            self.buf.insert(seq, data).is_none(),
+            "duplicate segment {}",
+            seq
+        );
```
marcbowes, line 183: I made a similar comment on a previous PR.

When I read this I skipped the implication of L180 because I don't read assertions until I understand the body of the method entirely. That part then got stuck on "but where do you insert?"..

I think it's worth an extra line of code to separate the work from the assertion.

src/host.rs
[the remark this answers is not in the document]
marcbowes: We should crash or log then. Or at least leave a comment.

marcbowes (approved):

README.md
mcches: yup

src/dns.rs
mcches: A bit, but the trait is opaque so I'm good to follow up on this.

src/dns.rs
mcches: done

src/envelope.rs
mcches: I took another pass on this.

src/host.rs
mcches: Good catch, added a test.

src/host.rs
mcches: We could limit, but then we need to panic if the buffer is full. The network delivers into these channels and without some form of back pressure / flow control I'm not seeing how this would work. I recall we discussed punting back pressure a few weeks back. Do you think we should get something in now?

src/host.rs
mcches: See ^.

src/host.rs
mcches: The _host has L4_ software really resonated with me :) I can use new instead if this is overboard.

src/host.rs
mcches: I'll stop using this pattern.

src/envelope.rs
mcches: I adding another layer of modeling to avoid this.

src/host.rs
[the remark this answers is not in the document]
davidbarsky: agreed with marc. i would crash here with an `unreachable!()`.

(also, hi marc!)

davidbarsky:
> for spans can those just be the instrument macro?

for some methods, yes. however, it feels a little odd to apply `#[instrument]`to `poll_`-based methods (i'm going off vibes and aesthetics more than anything) and in circumstances where we discover what IP address is used inside the method. in the latter case, while we can `.record()` those values to `Span::current()`, i don't think it's a good idea to this inside of a library, especially since the span returned by `Span::current()` might not be the _lexically scoped_ span. I'd make a span manually here instead, but I'll outline an approach on the issue @mcches made for spans.

src/host.rs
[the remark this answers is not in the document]
mcches: Check out the latest... I bounded the channels and I panic if either the receiver drops (bug) or the channel is at capacity. We can circle back on back pressure at a later date.

src/envelope.rs
davidbarsky: whoops, might've commented on a stable revision. sorry about that!

PR #51, opened 2022-10-28
TcpListener implements `Send`

mcches:
Interior mutability was initially chosen to align with the `tokio::net`
API. This actually diverged, as the listener could not be moved into
spawned tasks.

We go back to a simple `Notify` and `VecDeque` channel, which when Arc'd
can be moved into spawned tasks.

### DIFF
--- a/tests/async_send_sync.rs
+++ b/tests/async_send_sync.rs
@@ -1,0 +1,96 @@
+//! Copied over from:
+//! https://github.com/tokio-rs/tokio/blob/master/tokio/tests/async_send_sync.rs
+
+#[allow(dead_code)]
+fn require_send<T: Send>(_t: &T) {}
+#[allow(dead_code)]
+fn require_sync<T: Sync>(_t: &T) {}
+#[allow(dead_code)]
+fn require_unpin<T: Unpin>(_t: &T) {}
+
+#[allow(dead_code)]
+struct Invalid;
+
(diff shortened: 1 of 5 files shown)


tests/tcp.rs
```diff
@@ -175,9 +175,11 @@ fn accept_front_of_line_blocking() -> Result {
     sim.host("B", || async {
         let listener = bind().await;
 
-        while let Ok((_, peer)) = listener.accept().await {
-            tracing::debug!("peer {}", peer);
-        }
+        tokio::spawn(async move {
```
davidbarsky, line 178: this is probably a reasonable test, but i think you can also copy tokio's assert send/sync macro into turmoil and make these tests first-class: https://github.com/tokio-rs/tokio/blob/master/tokio/tests/async_send_sync.rs#L117

mcches: TIL, thanks for sharing!

tests/async_send_sync.rs
```diff
@@ -0,0 +1,96 @@
+//! Copied over from:
+//! https://github.com/tokio-rs/tokio/blob/master/tokio/tests/async_send_sync.rs
+
+#[allow(dead_code)]
```
davidbarsky, line 4: this is considered dead code? wild.

davidbarsky (approved):

Issue #49, opened 2022-10-25
Look into spans for network tracing


mcches:
See comments in #48 re: spans. 

For `TcpStream` and `UdpSocket` spans might simplify the context needed for each event, ie syn, fin, etc.

mcches:
I did an oops and this shows up in docs: https://docs.rs/turmoil/0.3.0/turmoil/macro.trace.html. No biggie, but we could hide it as part of this task.

PR #53, opened 2022-11-09
Use tracing levels for different network events

mcches:
Everything was at `info`, which got pretty noisy. We now use
`tracing` directly and leverage level to differentiate based on
event frequency.

Less frequent events (like new host or bind) use `info` and normal
network traffic (like send/hold/recv etc.) use `trace`.

Work still needs to be done here to integrate spans per #49.


src/host.rs
```diff
@@ -85,7 +85,7 @@ impl Host {
     pub(crate) fn receive_from_network(&mut self, envelope: Envelope) -> Result<(), Protocol> {
         let Envelope { src, dst, message } = envelope;
 
-        trace!(?dst, ?src, protocol = %message, "Delivered");
+        tracing::trace!(target: TRACING_TARGET, ?dst, ?src, protocol = %message, "Delivered");
```
LucioFranco, line 88: Isnt the target already set since the crate name is `turmoil`?

mcches: It is the full mod path, which isn't major but it's pretty nice to just see turmoil on all the events.

LucioFranco: Interesting, @davidbarsky I wonder what you think

mcches: Per #49 I assume everything that went to `trace` in this change would be a span. I don't like seeing the internal mod path in the traces, which is why I use a constant. It's not really a trace for the crate, rather it's a network trace that we happen to use tracing for. I don't feel too strongly though, just personal preference. Long term, I think we want a subscriber included that does nice things like write in different formats, and to different sinks like a file.

LucioFranco (approved):

Issue #28, opened 2022-09-21
Setup CI


mcches:
At min we should run a full build per PR.

See: https://github.com/jonhoo/rust-ci-conf

marcbowes:
We now run a build per PR, but there are some other things to do such as:

- setup build badges
- run clippy and/or other things

mcches:
Add badge: https://github.com/tokio-rs/turmoil/pull/33

mcches:
clippy is running, i'm going to close this.

PR #58, opened 2022-11-22
add clippy to CI

MarinPostma:
This PR adds Clippy to the CI

#28

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Git Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Rust Toolchain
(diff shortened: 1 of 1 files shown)


mcches:
Tagging #28 .

.github/workflows/rust.yml
```diff
@@ -31,3 +31,20 @@ jobs:
         with:
           command: test
           args: --verbose
+
+    clippy:
+        name: Run Clippy
```
mcches, line 36: [nit] indent looks like 4 instead of 2 here

.github/workflows/rust.yml
```diff
@@ -31,3 +31,20 @@ jobs:
...
+        runs-on: ${{ matrix.os }}
+        steps:
+          - uses: actions/checkout@v2
```
mcches, line 39: looks like v3 is available

PR #60, opened 2022-11-22
rename ToSocketAddr to ToSocketAddrs

MarinPostma:
Rename `ToSocketAddr` to `ToSocketAddrs` to make turmoil more compatible with tokio.

My motivation is that I am trying to mock parts of tokio with turmoil in my project, behind a feature flags. This is only possible is the crates are at least name-compatible.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -10,7 +10,7 @@ pub trait ToIpAddr {
     fn to_ip_addr(&self, dns: &mut Dns) -> IpAddr;
 }
 
-pub trait ToSocketAddr {
+pub trait ToSocketAddrs {
     fn to_socket_addr(&self, dns: &Dns) -> SocketAddr;
 }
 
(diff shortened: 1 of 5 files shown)


mcches:
The turmoil implementation is explicitly singleton v. plural like the tokio one. I didn't have issues integrating due to the supplied `From` conversions. One nice feature we have here is that human readable hostnames are supported, which makes debugging via the trace a lot easier. 

I'm not opposed since the trait is opaque anyways and we can do whatever we like internally.

Could you share an example snippet that was troublesome?

MarinPostma:
imagine a function like:
```rust
async fn setup_server(addrs: impl ToSocketAddrs) {
     let listener = TcpListener::bind(addrs).await;
//...
}
```

mcches:
> Consider a function like:
> 
> ```rust
> async fn setup_server(addrs: impl ToSocketAddrs) {
>      let listener = TcpListener::bind(addrs).await;
> //...
> }
> ```

I see. Given the nature of that trait I generally see a concrete addr passed in this use case and then the `From` conversions do their job.

I'm good with aligning the trait, but we have to bump the major version so it might be worth stashing this and getting some of the net gaps fleshed out with minor version releases if you can unblock yourself.

We also should export as `turmoil::net::ToSocketAddrs` to align.

mcches (approved):
lgtm!

PR #61, opened 2022-11-22
add poll_accept to tcp_stream

MarinPostma:
This PR adds the `poll_accept` method to `TcpStream`

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -319,13 +319,15 @@ impl Tcp {
         addr: SocketAddr,
         cx: &mut Context<'_>,
     ) -> Poll<(Syn, SocketAddr)> {
-        let socket = &mut self.binds[&addr];
-        match socket.deque.pop_front() {
-            Some(e) => Poll::Ready(e),
-            None => {
-                socket.waker.replace(cx.waker().clone());
-                Poll::Pending
-            }
+        match self.binds.get_mut(&addr) {
+            Some(socket) => match socket.deque.pop_front() {
+                Some(e) => Poll::Ready(e),
+                None => {
+                    socket.waker.replace(cx.waker().clone());
+                    Poll::Pending
+                }
+            },
+            None => panic!("no binding for {addr}"),
         }
     }
 
--- a/src/host.rs
+++ b/src/host.rs
@@ -187,7 +187,10 @@ pub(crate) struct Tcp {
 }
 
 struct ServerSocket {
-    /// Notify the TcpListener when SYNs are delivered
+    /// `Waker` from the `Context` of the last call to `poll_accept`.
+    ///
+    /// Similarly to tokio, only the `Waker` from the context of the last
+    /// call `poll_accept` is kept.
     waker: Option<Waker>,
 
     /// Pending connections for the TcpListener to accept
@@ -307,6 +310,10 @@ impl Tcp {
         rx
     }
 
+    /// Poll for new connections to `addr`.
+    ///
+    /// If there are none, the `Waker` from `cx` is stored, to wake up the task when
+    /// a new connection become available.
     pub(crate) fn poll_accept(
         &mut self,
         addr: SocketAddr,
--- a/src/host.rs
+++ b/src/host.rs
@@ -400,7 +400,35 @@ impl Tcp {
 
 #[cfg(test)]
 mod test {
+    use std::sync::atomic::{AtomicU8, Ordering};
+    use std::sync::Arc;
+    use std::task::{Context, Poll, Wake};
+
+    use futures::FutureExt;
+    use tokio::sync::oneshot;
+
+    use crate::envelope::{Segment, Syn};
     use crate::{Host, Result};
+
+    use super::Tcp;
+
+    struct MockWaker(AtomicU8);
+
+    impl MockWaker {
+        fn new() -> Self {
+            MockWaker(AtomicU8::new(0))
+        }
+
+        fn get(&self) -> u8 {
+            self.0.load(Ordering::Relaxed)
+        }
+    }
+
+    impl Wake for MockWaker {
+        fn wake(self: std::sync::Arc<Self>) {
+            self.0.fetch_add(1, Ordering::Relaxed);
+        }
+    }
 
     #[test]
     fn recycle_ports() -> Result {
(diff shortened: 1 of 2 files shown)


mcches:
Looks like some existing tests are failing due to timeouts when I ran them locally. It looks like returning `Pending` drops that future and we never wake.

mcches (changes requested):
See my top level comment about the test failures. I did something similar here where I stashed away the future across polls, but it wasn't pretty.

https://github.com/tokio-rs/turmoil/commit/28f09244cbbdec14ac225ccdd2370960ca8a6dc9#diff-820e47c99030d81ae1002778f18cb0a591d7e759f0334cd0c3e26435d4e02f51R57

e.g. https://docs.rs/tokio-util/latest/src/tokio_util/sync/poll_semaphore.rs.html#55

MarinPostma:
@mcches came up with a very similar hack to yours. Really feels like this is stretching the design very thin though...

mcches:
> @mcches came up with a very similar hack to yours. Really feels like this is stretching the design very thin though...
> 
> p.s: happy thanksgiving ;)

I agree this is stretching it. I think we should avoid the mutex if possible, but `TcpListener::accept` being immutable is problematic.

src/net/tcp/listener.rs
```diff
@@ -48,7 +60,26 @@ impl TcpListener {
...
+    /// Polls to accept a new incoming connection to this listener.
+    ///
+    /// If there is no connection to accept, Poll::Pending is returned and the current task will be
+    /// notified by a waker. Note that on multiple calls to poll_accept, only the Waker from the
+    /// Context passed to the most recent call is scheduled to receive a wakeup.
+    pub fn poll_accept(&self, cx: &mut Context<'_>) -> Poll<Result<(TcpStream, SocketAddr)>> {
         loop {
+            {
+                let mut lock = self.accept_fut.lock().unwrap();
+                ready!((*lock).poll(cx));
```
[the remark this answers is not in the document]
MarinPostma: we are polling the future we put there on line 77, if we don't nothing happens, and this would basically be a spin lock no?

src/net/tcp/listener.rs
```diff
@@ -48,7 +60,26 @@ impl TcpListener {
...
+
+    /// Polls to accept a new incoming connection to this listener.
+    ///
+    /// If there is no connection to accept, Poll::Pending is returned and the current task will be
+    /// notified by a waker. Note that on multiple calls to poll_accept, only the Waker from the
+    /// Context passed to the most recent call is scheduled to receive a wakeup.
+    pub fn poll_accept(&self, cx: &mut Context<'_>) -> Poll<Result<(TcpStream, SocketAddr)>> {
         loop {
+            {
+                let mut lock = self.accept_fut.lock().unwrap();
```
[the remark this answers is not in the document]
MarinPostma: Agreed. I think this ugly pattern will keep showing up, for basically all `poll_` implementation. Maybe we should go ahead and refactor the code to work nicely for poll_ functions, and then straightforwardly build the async version on top of them?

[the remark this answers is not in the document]
mcches: Two more ideas:

/1/ Use `unimplemented!(...)` for `poll_` methods. It'd be good to hear more about the use case that needs polling variants.

/2/ Stash the poll state in the host itself. This would avoid the need for a mutex. We can't take the above approach directly as we can't hold onto the "world" across an await, but we could potentially grab the waker from the cx passed to poll, stash it with an identifier for the listener, and then _wake_ once a syn arrives.

MarinPostma:
@mcches 

changed the approach, and got rid of the notifier. Instead, I used wakers directly, and this is a lot nicer.

mcches (commented):
Nice! I like how simple this ended up.

I think we have decent test coverage, but I'm wondering if the following is covered:

1/ poll_accept and set the waker
2/ syn arrives and we wake
3/ another syn arrives before we reset the waker

MarinPostma:
@mcches I have updated the docs, and added tests

src/host.rs
```diff
@@ -176,8 +176,11 @@ pub(crate) struct Tcp {
 }
 
 struct ServerSocket {
-    /// Notify the TcpListener when SYNs are delivered
-    notify: Arc<Notify>,
+    /// `Waker` from the `Context` of the last call to `poll_accept`.
+    ///
+    /// Similarly to tokio, only the `Waker` from the context of the last
+    /// call `poll_accept` is kept.
```
mcches, line 182:
```suggestion
    /// call to `poll_accept` is kept.
```

src/host.rs
```diff
@@ -297,8 +299,25 @@ impl Tcp {
         rx
     }
 
-    pub(crate) fn accept(&mut self, addr: SocketAddr) -> Option<(Syn, SocketAddr)> {
-        self.binds[&addr].deque.pop_front()
+    /// Poll for new connections to `addr`.
+    ///
+    /// If there are none, the `Waker` from `cx` is stored, to wake up the task when
+    /// a new connection become available.
```
mcches, line 305:
```suggestion
    /// a new connection becomes available.
```

src/host.rs
```diff
@@ -388,4 +437,54 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn poll_accept() {
```
mcches, line 442: Could we test this from the `TcpListener` level in the existing `tcp.rs` test? You wouldn't need to mock out anything then.

mcches (commented):
The approach looks good, just a few more comments.

src/host.rs
MarinPostma: @mcches, I just got back into it, and I'm not quite sure how this would be done. First, the `accept` method basically just calls the `poll_accept` method, so we are basically pretty well covered here. For the specific case we're testing here, we require a `Context`. If we're testing from the higher level `TcpStream`, we can get such a waker from `poll_fn`, but I'm not sure what we can do from there: we can't synchronize the client and the server from within  `poll_fn`, we can't inspect what happens with the waker, etc...

Maybe you see something that I don't?

mcches:
At this time I think we should invest in the path of moving turmoil into tokio rather than expand on the API surface, especially for more legacy methods like `poll_`. 

Is this blocking or can we come up with a temporary workaround?

mcches:
Closing due to no activity.

PR #63, opened 2022-12-04
Refactor stream socket

MarinPostma:
This PR refactors `StreamSocket` to make use of an internal buffer along with a waker instead of using a tokio channel. This is a step towards implementing the `peek` method family on TCP stream.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -3,13 +3,14 @@ use crate::net::{SocketPair, TcpListener, UdpSocket};
 use crate::world::World;
 use crate::{Envelope, TRACING_TARGET};
 
-use bytes::Bytes;
+use bytes::{Buf, Bytes};
 use indexmap::IndexMap;
 use std::collections::VecDeque;
 use std::fmt::Display;
 use std::io;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
+use std::task::{Context, Poll, Waker};
 use tokio::sync::{mpsc, Notify};
 use tokio::time::{Duration, Instant};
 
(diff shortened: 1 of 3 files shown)


src/net/tcp/stream.rs
```diff
@@ -115,48 +110,39 @@ impl TcpStream {
...
     pub(crate) pair: Arc<SocketPair>,
-    receiver: mpsc::Receiver<SequencedSegment>,
-    /// FIN received, EOF for reades
-    is_closed: bool,
 }
 
 impl ReadHalf {
     fn poll_read_priv(&mut self, cx: &mut Context<'_>, buf: &mut ReadBuf) -> Poll<Result<()>> {
-        if self.is_closed || buf.capacity() == 0 {
+        if buf.capacity() == 0 {
```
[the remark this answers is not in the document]
MarinPostma: The same effect is achieved by:
- retuning ready when closed: https://github.com/tokio-rs/turmoil/pull/63/files/2f2dc8552c0e6dce18556f094c54dce8147dfe1a#diff-f03a53510db113b4766b3ae24736c745a95ee30b9214eb509027fe6d609de868R230
- and then, try_read will return 0: https://github.com/tokio-rs/turmoil/pull/63/files/2f2dc8552c0e6dce18556f094c54dce8147dfe1a#diff-f03a53510db113b4766b3ae24736c745a95ee30b9214eb509027fe6d609de868R238

src/host.rs
```diff
@@ -388,4 +430,68 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn socket_read() {
```
mcches, line 435: Can we test what follows at the TcpStream level instead?

MarinPostma: pretty much the same issue as in https://github.com/tokio-rs/turmoil/pull/61#discussion_r1062734834
What exactly can we test at a higher level that is not already covered by the test we have, given that the only thing we can really do is wrap the poll methods in a `poll_fn`?

mcches:
Would this be simplified by #107 ?

mcches:
Closing due to no activity.

PR #65, opened 2022-12-06
Fix host elapsed time across software restarts

mcches:
Previously, we simply tracked the elapsed duration after each tick. This makes it impossible to get a host-level view of elapsed time from within the software. Software _could_ use an `Instant` + base `SystemTime` approach, but the `Instant` has issues across host restarts as the tokio runtime is replaced.

This change fixes that by tracking bulk elapsed time on each tick and uses that as a base along with a new `Instant` that is set each time the software runs. With the elapsed base we can accurately give a current elapsed time to running software and survive restarts.

I admit this isn't the cleanest, but I can't figure out how to use a single `Instant` across runtimes, and re-using the runtime seems a lot more complicated.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,7 @@
 use crate::*;
 
 use rand::{RngCore, SeedableRng};
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 
 /// Configure the simulation
 pub struct Builder {
@@ -22,6 +22,12 @@ impl Builder {
                 message_loss: Some(config::MessageLoss::default()),
             },
         }
+    }
+
+    /// When the simulation starts.
+    pub fn epoch(&mut self, value: SystemTime) -> &mut Self {
+        self.config.epoch = value;
+        self
     }
 
     /// How long the test should run for in simulated time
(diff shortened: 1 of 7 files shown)


mcches:
I have a test locally, but I'll wait for buy-in before proceeding with this direction.

mcches:
Here is the gist...

```rust
#[derive(Clone)]
struct WallClock {
    since_epoch: Duration,
}

impl WallClock {
    fn now(&self) -> u128 {
        self.since_epoch.as_nanos() + elapsed().as_nanos()
    }
}

#[test]
fn elapsed_time_across_restarts() -> Result {
    let mut sim = Builder::new()
        .tick_duration(Duration::from_millis(6))
        .build();

    let clock = WallClock {
        since_epoch: sim.since_epoch(),
    };

    sim.host("host", move || {
        let clock = clock.clone();
        println!("start!");

        async move {
            loop {
                println!("now {:?}", clock.now());
                tokio::time::sleep(Duration::from_millis(2)).await;
            }
        }
    });

    sim.run()?;

    sim.bounce("host");
    sim.run()
}
```

```txt
start!
now 1670301697962301000
now 1670301697964301000
now 1670301697966301000
start!
now 1670301697968301000
now 1670301697970301000
now 1670301697972301000
```

carllerche (approved):
LGTM, thanks

PR #67, opened 2022-12-09
Add host error support

battesonb:
This change only captures a host failure during a run of a simulation. Bounces assign a new JoinHandle which is checked after a subsequent run. This change is a breaking change for those that do not produce a `Result` or loop forever.

Since this approach removes the host runtime from the `rts` map when it completes, a few of the existing tests would fail as all of the resources owned by completed hosts would be dropped. I opted to use `future::pending()` to prevent the host from finishing on the happy path, as this seems preferable. However, I can handle hosts differently if the previous behavior is preferred.

This addresses #36.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ let mut sim = turmoil::Builder::new().build();
 sim.host("server", || async move {
     // host software goes here
 
-    Err("Host software finished unexpectedly")?
+    Ok(())
 });
 
 // register a client
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ let mut sim = turmoil::Builder::new().build();
 // register a host
 sim.host("server", || async move {
     // host software goes here
+
+    Err("Host software finished unexpectedly")?
 });
 
 // register a client
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -88,7 +88,7 @@ impl<'a> Sim<'a> {
     pub fn host<F, Fut>(&mut self, addr: impl ToIpAddr, host: F)
     where
         F: Fn() -> Fut + 'a,
-        Fut: Future<Output = ()> + 'static,
+        Fut: Future<Output = Result> + 'static,
     {
         let rt = Rt::new();
         let addr = self.lookup(addr);
@@ -100,13 +100,9 @@ impl<'a> Sim<'a> {
             world.register(addr);
         }
 
-        World::enter(&self.world, || {
-            rt.with(|| {
-                tokio::task::spawn_local(host());
-            });
-        });
+        let handle = World::enter(&self.world, || rt.with(|| tokio::task::spawn_local(host())));
 
-        self.rts.insert(addr, Role::simulated(rt, host));
+        self.rts.insert(addr, Role::simulated(rt, host, handle));
     }
 
     /// Crash a host. Nothing will be running on the host after this method. You
@@ -117,18 +113,22 @@ impl<'a> Sim<'a> {
             Role::Simulated { rt, .. } => {
                 rt.cancel_tasks();
             }
-        });
+        })
     }
 
     /// Bounce a host. The software is restarted.
     pub fn bounce(&mut self, addr: impl ToIpAddr) {
         self.run_with_host(addr, |rt| match rt {
             Role::Client { .. } => panic!("can only bounce hosts, not clients"),
-            Role::Simulated { rt, software } => {
+            Role::Simulated {
+                rt,
+                software,
+                handle,
+            } => {
                 rt.cancel_tasks();
-                rt.with(|| tokio::task::spawn_local(software()));
+                handle.replace(rt.with(|| tokio::task::spawn_local(software())));
             }
-        });
+        })
     }
 
     // Run `f` with the host at `addr` set on the world.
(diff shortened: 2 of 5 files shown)


src/sim.rs
```diff
@@ -245,15 +245,31 @@ impl<'a> Sim<'a> {
...
             // this stage.
             for addr in finished.iter() {
-                if let Some(Role::Client { rt, handle }) = self.rts.remove(addr) {
-                    rt.block_on(handle)??;
+                if let Some(role) = self.rts.remove(addr) {
+                    let (rt, handle) = match role {
+                        Role::Client { rt, handle } => (rt, Some(handle)),
+                        Role::Simulated { rt, handle, .. } => (rt, handle),
+                    };
+                    if let Some(handle) = handle {
```
LucioFranco, line 270: I wonder if we want to log here to allow people to identify when their rt isn't shutting down gracefully? Like what task is blocking it etc.

src/sim.rs
```diff
@@ -282,6 +298,7 @@ mod test {
         time::Duration,
     };
 
+    use futures::future;
```
LucioFranco, line 301: Should use the ones from std https://doc.rust-lang.org/std/future/fn.pending.html

tests/tcp.rs
```diff
@@ -5,6 +5,7 @@ use std::{
     time::Duration,
 };
 
+use futures::future;
```
LucioFranco, line 8: where is this used?

LucioFranco (commented):
Overall seems like a good addition, left a few comments/questions

tests/tcp.rs
battesonb: Lines 340 and 391 of this file, but I will replace with the std import as you called out above.

src/sim.rs
battesonb: Perhaps! As it is, the tests that return results just print out the error, but not the source host/client and the backtrace does not go further than the test entry point.

tests/tcp.rs
LucioFranco: oh i misread this anyways thought this was not a test file my bad

src/sim.rs
mcches: I think this will end up double logging. You'll see the turmoil warn if you have tracing enabled and then the test will render the error on failure. As an author of host software I expect my entry point to have proper error handling and logging so maybe we don't need to provide that here.

README.md
```diff
@@ -40,6 +40,8 @@ let mut sim = turmoil::Builder::new().build();
 // register a host
 sim.host("server", || async move {
     // host software goes here
+
+    Err("Host software finished unexpectedly")?
```
mcches, line 44: `Ok(())`?

src/sim.rs
```diff
@@ -245,15 +245,36 @@ impl<'a> Sim<'a> {
                     }
                     is_finished = is_finished && handle.is_finished();
                 }
+
+                if let Role::Simulated {
```
mcches, line 249: Perhaps a match now that both variants are touched?

mcches (commented):
Looks pretty good, thanks for picking up this change!

We haven't really come across the version issue yet, so I'm not sure if we should cut a branch for v0.4 or just go to mainline and the next release is a bump. @LucioFranco thoughts?

> a few of the existing tests would fail as all of the resources owned by completed hosts would be dropped.

It's not clear to me why the non-terminal host change is needed. What is dropped that fails the test?

README.md
battesonb: Yes, wanted to show a good default since it's almost always the case that it's undesired behavior, unless you're testing a shutdown. But will switch to `Ok(())` for brevity.

battesonb:
> It's not clear to me why the non-terminal host change is needed. What is dropped that fails the test?

You can see a few of the tests that have had the `future::pending()` call added have this problem. So for example, the `crash_blocks_until_complete` test went from having its host role held in the `rts` map to being marked as finished and thus being removed from the map. This results in the spawned tokio task that was blocked by the 0-semaphore being dropped and the `Arc`s reference count decreasing by one during the `sim.run()?` call and failing the first assertion.

LucioFranco:
> We haven't really come across the version issue yet, so I'm not sure if we should cut a branch for v0.4 or just go to mainline and the next release is a bump. @LucioFranco thoughts?

imo if there is breaking code it should just be released as a new breaking change, its pretty low cost to do so.

mcches:
> > It's not clear to me why the non-terminal host change is needed. What is dropped that fails the test?
> 
> You can see a few of the tests that have had the `future::pending()` call added have this problem. So for example, the `crash_blocks_until_complete` test went from having its host role held in the `rts` map to being marked as finished and thus being removed from the map. This results in the spawned tokio task that was blocked by the 0-semaphore being dropped and the `Arc`s reference count decreasing by one during the `sim.run()?` call and failing the first assertion.

It makes sense to me in `crash_blocks_until_complete`, just not sure about _all_ usages.

battesonb:
> > > It's not clear to me why the non-terminal host change is needed. What is dropped that fails the test?
> > 
> > You can see a few of the tests that have had the `future::pending()` call added have this problem. So for example, the `crash_blocks_until_complete` test went from having its host role held in the `rts` map to being marked as finished and thus being removed from the map. This results in the spawned tokio task that was blocked by the 0-semaphore being dropped and the `Arc`s reference count decreasing by one during the `sim.run()?` call and failing the first assertion.
> 
> It makes sense to me in `crash_blocks_until_complete`, just not sure about _all_ usages.

Oh it's definitely not required for all of them. `Ok(())` or an error are perfect for a host that has software that is intended to complete. In the case of the turmoil tests, that's most of them. Otherwise, looping forever to accept and handle connections is a reasonable approach.

I wouldn't expect consumers of turmoil to use the `future::pending()` call for most practical uses.

mcches (approved):
Lgtm

mcches:
I just thought of one edge case...

Is it possible to bounce a host mid run and wipe out a host software failure?

PR #71, opened 2023-01-05
Add more type conversions for ToSocketAddrs

LucioFranco:


src/dns.rs
```diff
@@ -88,3 +90,28 @@ impl ToSocketAddrs for (IpAddr, u16) {
...
+    fn to_socket_addr(&self, _: &Dns) -> SocketAddr {
+        self.parse().unwrap()
+    }
+}
+
+impl ToSocketAddrs for String {
+    fn to_socket_addr(&self, _: &Dns) -> SocketAddr {
+        self.parse().unwrap()
+    }
+}
```
MarinPostma, line 110: shouldn't we be doing some DNS lookups here like in https://github.com/tokio-rs/turmoil/blob/main/src/dns.rs#L65-L90?

LucioFranco: oh yes I think I copied the tokio code oops

LucioFranco: wait no some of them don't go through dns and some do. We will need to fix that.

mcches (approved):
str parse will be nice to have

MarinPostma (approved):

PR #77, opened 2023-01-19
Add option to query hosts via regex

battesonb:
This allows various host interactions via regex matches to reduce the effort to induce more complex interactions. While this has changed the explicit API contract in some cases, this should not affect most test cases as the `ToIpAddrs` trait implements `ToIpAddr` for all `T` that implement `ToIpAddr`.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -81,9 +81,12 @@ where
     }
 }
 
-impl ToIpAddrs for Vec<IpAddr> {
-    fn to_ip_addrs(&self, _: &mut Dns) -> Vec<IpAddr> {
-        self.clone()
+impl<T> ToIpAddrs for Vec<T>
+where
+    T: ToIpAddr,
+{
+    fn to_ip_addrs(&self, dns: &mut Dns) -> Vec<IpAddr> {
+        self.iter().map(|addr| addr.to_ip_addr(dns)).collect()
     }
 }
 
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -82,15 +82,6 @@ where
     }
 }
 
-impl<T> ToIpAddrs for Vec<T>
-where
-    T: ToIpAddr,
-{
-    fn to_ip_addrs(&self, dns: &mut Dns) -> Vec<IpAddr> {
-        self.iter().map(|addr| addr.to_ip_addr(dns)).collect()
-    }
-}
-
 #[cfg(feature = "regex")]
 impl ToIpAddrs for Regex {
     fn to_ip_addrs(&self, dns: &mut Dns) -> Vec<IpAddr> {
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,4 +1,5 @@
 use indexmap::IndexMap;
+#[cfg(feature = "regex")]
 use regex::Regex;
 use std::net::{IpAddr, SocketAddr};
 
@@ -90,6 +91,7 @@ where
     }
 }
 
+#[cfg(feature = "regex")]
 impl ToIpAddrs for Regex {
     fn to_ip_addrs(&self, dns: &mut Dns) -> Vec<IpAddr> {
         #[allow(clippy::needless_collect)]
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,4 +1,5 @@
 use indexmap::IndexMap;
+use regex::Regex;
 use std::net::{IpAddr, SocketAddr};
 
 pub struct Dns {
@@ -8,6 +9,10 @@ pub struct Dns {
 
 pub trait ToIpAddr {
     fn to_ip_addr(&self, dns: &mut Dns) -> IpAddr;
+}
+
+pub trait ToIpAddrs {
+    fn to_ip_addrs(&self, dns: &mut Dns) -> Vec<IpAddr>;
 }
 
 /// A simulated version of `tokio::net::ToSocketAddrs`.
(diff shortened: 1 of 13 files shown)


src/dns.rs
```diff
@@ -10,6 +11,10 @@ pub trait ToIpAddr {
     fn to_ip_addr(&self, dns: &mut Dns) -> IpAddr;
 }
 
+pub trait ToIpAddrs {
```
mcches, line 14: If we break the contract and need to bump the version anyways, should we consider consolidating into a single trait?

src/dns.rs
```diff
@@ -63,6 +72,32 @@ impl ToIpAddr for IpAddr {
...
+impl ToIpAddrs for Vec<IpAddr> {
+    fn to_ip_addrs(&self, _: &mut Dns) -> Vec<IpAddr> {
+        self.clone()
+    }
+}
+
+impl ToIpAddrs for Regex {
+    fn to_ip_addrs(&self, dns: &mut Dns) -> Vec<IpAddr> {
+        #[allow(clippy::needless_collect)]
+        let hosts = dns.names.keys().cloned().collect::<Vec<_>>();
```
mcches, line 93: Hmm, because of the `&mut`?

src/lib.rs
```diff
@@ -43,59 +43,85 @@ use world::World;
...
 /// Must be called from within a Turmoil simulation.
 pub fn lookup(addr: impl ToIpAddr) -> IpAddr {
     World::current(|world| world.lookup(addr))
 }
 
-/// Hold messages two hosts, until [`release`] is called.
+/// Lookup an IP address by host name. Use regex to match a number of hosts.
+///
+/// Must be called from within a Turmoil simulation.
+pub fn lookup_many(addr: impl ToIpAddrs) -> Vec<IpAddr> {
```
mcches, line 66: Seems only a Regex would be useful here, so perhaps we can just be explicit instead of using the trait?

src/sim.rs
```diff
@@ -105,20 +105,20 @@ impl<'a> Sim<'a> {
         self.rts.insert(addr, Role::simulated(rt, host, handle));
     }
 
-    /// Crash a host. Nothing will be running on the host after this method. You
-    /// can use [`Sim::bounce`] to start the host up again.
-    pub fn crash(&mut self, addr: impl ToIpAddr) {
-        self.run_with_host(addr, |rt| match rt {
+    /// Crashes a host(s). Nothing will be running on the matched hosts after this method. You can
```
mcches, line 108: Howe about _Crash/Bounce the resolved hosts_ for these docs?

src/sim.rs
```diff
@@ -131,14 +131,16 @@ impl<'a> Sim<'a> {
         });
     }
 
-    // Run `f` with the host at `addr` set on the world.
-    fn run_with_host(&mut self, addr: impl ToIpAddr, f: impl FnOnce(&mut Role)) {
-        let h = self.world.borrow_mut().lookup(addr);
-        let rt = self.rts.get_mut(&h).expect("missing host");
+    /// Run `f` with the host(s) at `addr` set on the world.
```
mcches, line 134: at `addr` is stale

src/sim.rs
```diff
@@ -148,6 +150,11 @@ impl<'a> Sim<'a> {
         self.world.borrow_mut().lookup(addr)
     }
 
+    /// Lookup an ip address by host name.
```
mcches, line 153: Comment is singular and code is plural, but also see comment above about only having the regex path.

src/sim.rs
```diff
@@ -479,4 +491,29 @@ mod test {
...
+                async move {
+                    count.fetch_add(1, Ordering::SeqCst);
+                    future::pending().await
+                }
+            });
+        }
+
+        sim.run()?;
+        assert_eq!(count.load(Ordering::SeqCst), 3);
+        sim.bounce(Regex::new("host-\\d").unwrap());
```
mcches, line 513: Consider matching a subset for this test?

mcches, line 513: Should be able to just `?` the regex constructor with how `Result` is defined.

mcches (commented):
Looks pretty good, just wondering if/how we want to break the current public contracts.

src/dns.rs
battesonb: We could do, however there is the need for a trait like `ToIpAddr` for the `sim#client` and `sim#host` functions. As in these cases, the trait doubles as a generator rather than only a matcher.

src/dns.rs
battesonb: Yes, I battled with this one for quite a while. There's an open issue for clippy not to produce a false positive in this situation: https://github.com/rust-lang/rust-clippy/issues/6066

src/lib.rs
battesonb: The `ToIpAddrs` trait could also support a Vec of string matches (actually not with this revision, but I'll update it), so for example another valid input would be:

```rust
lookup_many(vec!["host-a", "host-b"]);
```

I was also looking into passing predicate callbacks, but I ran into trouble with conflicting generics with the `ToIpAddr` impl for `ToIpAddrs`.

Changing this, however, will affect all the other methods that currently take `ToIpAddr` (hold, release, partition). It may be a nice convenience if we asserted that the `lookup_many` should produce at least one element in each set, as its use could otherwise produce subtle bugs.

LucioFranco: > It may be a nice convenience if we asserted that the lookup_many should produce at least one element in each set, as its use could otherwise produce subtle bugs.

Does this mean that if we look up two hosts we must see at least two hosts returning?

LucioFranco (approved):
This seems like a good addition 👍. I think its fine atm to do small breaking changes often since this is just a library for writing tests.

src/rt.rs
```diff
@@ -78,7 +78,7 @@ impl Rt {
         let (tokio, local) = init();
 
         _ = mem::replace(&mut self.tokio, tokio);
-        _ = mem::replace(&mut self.local, local);
+        drop(mem::replace(&mut self.local, local));
```
mcches, line 81: clippy lint?

mcches (approved):
Nice! This looks solid.

PR #78, opened 2023-01-19
Add network topology introspection

mcches:
The sim now supports iterating each host pair's network link, as well
as each message that is currently "in flight" on the network.
    
Link and message metadata is exposed. We can also now force delivery
of messages.
    
In combination with the step capabilities that were recently added we
now have the ability to run more complex test scenarios.

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -10,14 +10,16 @@ pub(crate) struct Envelope {
     pub(crate) message: Protocol,
 }
 
+/// Supported network protocols.
 #[derive(Debug)]
-pub(crate) enum Protocol {
+pub enum Protocol {
     Tcp(Segment),
     Udp(Datagram),
 }
 
+/// UDP datagram.
 #[derive(Debug)]
-pub(crate) struct Datagram(pub Bytes);
+pub struct Datagram(pub Bytes);
 
 /// This is a simplification of real TCP.
 ///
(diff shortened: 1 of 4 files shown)


src/envelope.rs
```diff
@@ -10,30 +10,32 @@ pub(crate) struct Envelope {
...
+pub enum Segment {
     Syn(Syn),
     Data(u64, Bytes),
     Fin(u64),
     Rst,
 }
 
 #[derive(Debug)]
-pub(crate) struct Syn {
+pub struct Syn {
```
mh32, line 38: Is there any reason you prefer `pub(crate)` to `pub` with visibility limited to crate at the module declaration?

src/sim.rs
```diff
@@ -192,6 +204,13 @@ impl<'a> Sim<'a> {
         world.topology.set_link_fail_rate(a, b, value);
     }
 
+    /// Access a [`LinksIter`] to introspect inflight messages between hosts.
+    pub fn links(&self, f: impl FnOnce(LinksIter)) {
```
mh32, line 208: Did you consider returning an iterator? Seems slightly more idiomatic

mh32 (approved):

src/envelope.rs
mcches: No reason, just started out this way. We could go back and and fix this at some point for the internals.

src/sim.rs
mcches: It would certainly be nicer, but I don't believe we can since we need to borrow the world's ref cell.

PR #81, opened 2023-02-09
Add `gRPC` example

LucioFranco:
This commit adds a simple gRPC example using tonic and hyper.

### DIFF
--- a/examples/grpc/helloworld.proto
+++ b/examples/grpc/helloworld.proto
@@ -1,0 +1,37 @@
+// Copyright 2015 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "io.grpc.examples.helloworld";
+option java_outer_classname = "HelloWorldProto";
+
+package helloworld;
+
+// The greeting service definition.
+service Greeter {
+  // Sends a greeting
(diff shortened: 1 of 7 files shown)


examples/grpc/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+        request: Request<HelloRequest>,
+    ) -> Result<Response<HelloReply>, Status> {
+        let reply = HelloReply {
+            message: format!("Hello {}!", request.into_inner().name),
+        };
+        Ok(Response::new(reply))
+    }
+}
+
+mod connector {
```
mcches, line 97: What do you think about a hyper feature for this in turmoil? Premature maybe.

examples/grpc/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    let addr = (IpAddr::from(Ipv4Addr::UNSPECIFIED), 9999);
+
+    let mut sim = Builder::new().build();
+
+    let greeter = GreeterServer::new(MyGreeter {});
+
+    sim.host("server", move || {
+        let greeter = greeter.clone();
+        async move {
+            let listener = net::TcpListener::bind(addr).await.unwrap();
```
mcches, line 36: You can avoid the unwraps in the server now, hosts now also fail the sim if they err.

examples/grpc/helloworld.proto
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "io.grpc.examples.helloworld";
+option java_outer_classname = "HelloWorldProto";
```
mcches, line 19: Can we remove all this?

mcches (approved):
Real nice!

examples/grpc/helloworld.proto
LucioFranco: Yeah probably

examples/grpc/src/main.rs
LucioFranco: I don't think the connection trait (the main reason we have to newtype) will be in hyper 1.0 so I am not sure if that is the right approach. Maybe adding a `turmoil-hyper` library could work.

Issue #79, opened 2023-02-03
Support loopback


mcches:
Hosts currently may only bind to `0.0.0.0`.

https://github.com/tokio-rs/turmoil/blob/main/src/net/tcp/listener.rs#L28

Add support to bind `127.0.0.1` to unblock loopback scenarios. We need to decide how network topology is affected by these changes. For example, it doesn't make sense to allow partitions within a host.

th7nder:
Hey! I'd love to pick this one up as my first issue.

Let's clarify, what are the loopback scenarios you mentioned?

mcches:
Allow hosts to bind to localhost (e.g. 127.0.0.1) socket addresses, which are only reachable from within said host.

th7nder:
```
#[test]
fn connects_within_a_host() -> Result {

    let mut sim = Builder::new().build();

    sim.host("server", || async {
        let listener = TcpListener::bind((IpAddr::from(Ipv4Addr::LOCALHOST), PORT)).await?;

        let _ = listener.accept().await;
        tokio::spawn(async move {
            loop {
                if let Ok((mut stream, _socket)) = listener.accept().await {
                    let _ = stream.write_u8(7).await;
                }
            }
        });

        let mut stream = TcpStream::connect(("server", PORT))
            .await
            .expect("should have connected");

        assert_eq!(7, stream.read_u8().await.expect("successful read"));

        Ok(())
    });

    sim.client("client", async {
        assert_error_kind(
            TcpStream::connect(("server", PORT)).await,
            io::ErrorKind::ConnectionRefused,
        );
        Ok(())
    });
}
```
Questions:
1. Am I going into the right direction?
2. Should we consider being able to use `127.0.0.1` or `localhost` inside of a host as a reference to itself? I noticed that DNS uses 127.0.x.y addresses internally.

mcches:
> Am I going into the right direction?

Yes! I am curious why you need the first accept that ignores the result?

```rust
let _ = listener.accept().await;
```

> Should we consider being able to use 127.0.0.1 or localhost inside of a host as a reference to itself? I noticed that DNS uses 127.0.x.y addresses internally.

I think that makes a lot of sense. We should probably reserve that IP adds and also make sure `"localhost"` resolves to it in DNS.

So when you connect from within a host you can just do:

```rust
let mut stream = TcpStream::connect(("localhost", PORT)).await?;
```

th7nder:
[...]

PR #83, opened 2023-02-11
Add support for localhost binding in TCP (#79)

th7nder:
Implementation proposal for #79 .
It does not use topology under the hood, skips the chain.
The messages are not queued like in the topology, but are delivered to hosts directly.

If the approach is acceptable, I can add an another CR for UDP.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,9 +1,9 @@
 use indexmap::IndexMap;
 #[cfg(feature = "regex")]
 use regex::Regex;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr, Ipv6Addr};
 
-/// Each new registered host is thought to have an IP in a subnet 192.168.0.0/24
+/// Each new registered host has an IP in a subnet 192.168.0.0/24
 /// This is just a choice.
 /// localhost always resolves to 127.0.0.1, there is no localhost implementation for IPv6.
 pub struct Dns {
@@ -114,6 +114,10 @@ impl<'a> ToSocketAddrs for (&'a str, u16) {
             return (ip, self.1).into();
         }
 
+        if let Ok(ip) = self.0.parse::<Ipv6Addr>() {
+            return (ip, self.1).into();
+        }
+
         match dns.names.get(self.0) {
             Some(ip) => (*ip, self.1).into(),
             None => panic!("no ip address found for a hostname"),
@@ -186,12 +190,14 @@ mod tests {
     #[test]
     fn parse_str() {
         let mut dns = Dns::new();
-        dns.names.insert("foo".into(), "127.0.0.1".parse().unwrap());
-        let hostname_port = "foo:5000".to_socket_addr(&dns);
-        let ip_port = "127.0.0.1:5000".to_socket_addr(&dns);
+        let generated_addr = dns.lookup("foo");
 
-        let expected = "127.0.0.1:5000".parse().unwrap();
-        assert_eq!(hostname_port, expected);
-        assert_eq!(ip_port, expected);
+        let hostname_port = "foo:5000".to_socket_addr(&dns);
+        let ipv4_port = "127.0.0.1:5000";
+        let ipv6_port = "[::1]:5000";
+
+        assert_eq!(hostname_port, format!("{}:5000", generated_addr).parse().unwrap());
+        assert_eq!(ipv4_port.to_socket_addr(&dns), ipv4_port.parse().unwrap());
+        assert_eq!(ipv6_port.to_socket_addr(&dns), ipv6_port.parse().unwrap());
     }
 }
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,11 +1,15 @@
 use indexmap::IndexMap;
 #[cfg(feature = "regex")]
 use regex::Regex;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr, Ipv6Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 
 /// Each new registered host has an IP in a subnet 192.168.0.0/24
 /// This is just a choice.
-/// localhost always resolves to 127.0.0.1, there is no localhost implementation for IPv6.
+///
+/// Default mappings:
+/// localhost -> 127.0.0.1
+/// ip6-localhost -> ::1
+/// ip6-loopback -> ::1
 pub struct Dns {
     next: u16,
     names: IndexMap<String, IpAddr>,
@@ -30,7 +34,15 @@ impl Dns {
         let mut names: IndexMap<String, IpAddr> = IndexMap::new();
         names.insert(
             "localhost".to_string(),
-            std::net::Ipv4Addr::LOCALHOST.into(),
+            Ipv4Addr::LOCALHOST.into(),
+        );
+        names.insert(
+            "ip6-localhost".to_string(),
+            Ipv6Addr::LOCALHOST.into(),
+        );
+        names.insert(
+            "ip6-loopback".to_string(),
+            Ipv6Addr::LOCALHOST.into(),
         );
 
         Dns { next: 1, names }
@@ -109,18 +121,14 @@ impl ToSocketAddrs for (String, u16) {
 
 impl<'a> ToSocketAddrs for (&'a str, u16) {
     fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
-        // When "x.y.a.b" is passed as a hostname.
-        if let Ok(ip) = self.0.parse::<Ipv4Addr>() {
-            return (ip, self.1).into();
-        }
-
-        if let Ok(ip) = self.0.parse::<Ipv6Addr>() {
+        // When IP address is passed directly as a hostname.
+        if let Ok(ip) = self.0.parse::<IpAddr>() {
             return (ip, self.1).into();
         }
 
         match dns.names.get(self.0) {
             Some(ip) => (*ip, self.1).into(),
-            None => panic!("no ip address found for a hostname"),
+            None => panic!("no ip address found for a hostname: {}", self.0),
         }
     }
 }
@@ -185,6 +193,8 @@ mod sealed {
 
 #[cfg(test)]
 mod tests {
+    use std::net::{Ipv4Addr, Ipv6Addr};
+
     use crate::{dns::Dns, ToSocketAddrs};
 
     #[test]
@@ -196,8 +206,20 @@ mod tests {
         let ipv4_port = "127.0.0.1:5000";
         let ipv6_port = "[::1]:5000";
 
-        assert_eq!(hostname_port, format!("{}:5000", generated_addr).parse().unwrap());
+        assert_eq!(
+            hostname_port,
+            format!("{}:5000", generated_addr).parse().unwrap()
+        );
         assert_eq!(ipv4_port.to_socket_addr(&dns), ipv4_port.parse().unwrap());
         assert_eq!(ipv6_port.to_socket_addr(&dns), ipv6_port.parse().unwrap());
+    }
+
+    #[test]
+    fn localhosts() {
+        let mut dns = Dns::new();
+
+        assert_eq!(Ipv4Addr::LOCALHOST, dns.lookup("localhost"));
+        assert_eq!(Ipv6Addr::LOCALHOST, dns.lookup("ip6-localhost"));
+        assert_eq!(Ipv6Addr::LOCALHOST, dns.lookup("ip6-loopback"));
     }
 }
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,8 +1,11 @@
 use indexmap::IndexMap;
 #[cfg(feature = "regex")]
 use regex::Regex;
-use std::net::{IpAddr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
+/// Each new registered host is thought to have an IP in a subnet 192.168.0.0/24
+/// This is just a choice.
+/// localhost always resolves to 127.0.0.1, there is no localhost implementation for IPv6.
 pub struct Dns {
     next: u16,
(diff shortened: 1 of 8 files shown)


src/net/tcp/listener.rs
```diff
@@ -25,18 +29,25 @@ impl TcpListener {
     ///
     /// The returned listener is ready for accepting connections.
     ///
-    /// Only 0.0.0.0 is currently supported.
+    /// If you bind to the 0.0.0.0, you're effectivly binding to the generated IP address of the host.
+    /// See the Dns class for the current IP assignment.
```
mcches, line 33: Rust doesn't have classes and this is public documentation, which doesn't know about `Dns`.

src/dns.rs
```diff
@@ -103,9 +109,14 @@ impl ToSocketAddrs for (String, u16) {
 
 impl<'a> ToSocketAddrs for (&'a str, u16) {
     fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
+        // When "x.y.a.b" is passed as a hostname.
```
mcches, line 112: Nice catch.

src/dns.rs
```diff
@@ -176,8 +187,11 @@ mod tests {
     fn parse_str() {
         let mut dns = Dns::new();
         dns.names.insert("foo".into(), "127.0.0.1".parse().unwrap());
```
mcches, line 189: Can we change this to not be `127.0.0.1` and use lookup to generate an ip per your changes?

src/net/tcp/listener.rs
```diff
@@ -25,18 +29,25 @@ impl TcpListener {
...
+    ///
+    /// You can bind to `localhost`, which translates to 127.0.0.1.
+    /// It allows for the TCP socket to be only visible within a host and reachable *only* via localhost IPv4/IPv6 addresses.
     pub async fn bind<A: ToSocketAddrs>(addr: A) -> Result<TcpListener> {
```
mcches, line 37: What happens if someone binds to [`::1`](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#associatedconstant.LOCALHOST)?

src/world.rs
```diff
@@ -110,11 +110,24 @@ impl World {
         self.hosts.insert(addr, Host::new(addr));
     }
 
-    /// Send `message` from `src` to `dst`. Delivery is asynchronous and not
-    /// guaranteed.
+    /// Send `message` from `src` to `dst`.
+    /// Delivery between hosts is asynchronous and not guaranteed.
+    /// localhost communication does not use a topology.
     pub(crate) fn send_message(&mut self, src: SocketAddr, dst: SocketAddr, message: Protocol) {
```
mcches, line 116: As discussed offline let's see how it would look to send via the normal code path in top, but only support "drop" manipulation.

tests/tcp.rs
```diff
@@ -26,6 +26,68 @@ async fn bind() -> std::result::Result<TcpListener, std::io::Error> {
     TcpListener::bind((IpAddr::from(Ipv4Addr::UNSPECIFIED), PORT)).await
 }
 
+#[test]
+fn connects_within_a_localhost() -> Result {
+    let mut sim = Builder::new().build();
+
+    sim.host("server", || async {
```
mcches, line 33: You could simplify this by using `sim.client` and that is guaranteed to run to completion on `sim.run()`.

mcches (changes requested):
Thanks for picking up this change.

src/net/tcp/listener.rs
th7nder: It should work - I use methods which are IPv4 agnostic, besides the DNS. 
I will see.

th7nder: I made binding to ::1 separate.
You need to explicitly bind to ::1 to allow connections and it's separate stream than IPv4 localhost.

src/world.rs
th7nder: Made it happen.

src/top.rs
```diff
@@ -130,6 +150,27 @@ impl<'a> Iterator for LinkIter<'a> {
     }
 }
 
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
```
LucioFranco, line 153: Would be nice to have some docs here

src/top.rs
```diff
@@ -130,6 +150,27 @@ impl<'a> Iterator for LinkIter<'a> {
...
+
+impl Target {
+    fn new(src: IpAddr, dst: IpAddr) -> Target {
+        if src.is_loopback() {
+            Target(dst, Interface::Localhost)
+        } else if dst.is_loopback() {
+            Target(src, Interface::Localhost)
+        } else {
+            Target(dst, Interface::External)
+        }
```
LucioFranco, line 170: Would be nice to have some docs explaining the logic here

src/world.rs
```diff
@@ -101,6 +101,12 @@ impl World {
 
         tracing::info!(target: TRACING_TARGET, hostname = ?self.dns.reverse(addr), ?addr, "New");
 
+        // // Handles connection within a host
```
LucioFranco, line 104:
```suggestion
        // Handles connection within a host
```

src/top.rs
th7nder: Wonderful feedback. I just realized we don't need this type.

src/top.rs
th7nder: Wonderful feedback. I just realized we don't need this type.

src/dns.rs
```diff
@@ -103,9 +109,18 @@ impl ToSocketAddrs for (String, u16) {
...
+        }
+
+        if let Ok(ip) = self.0.parse::<Ipv6Addr>() {
+            return (ip, self.1).into();
+        }
+
         match dns.names.get(self.0) {
             Some(ip) => (*ip, self.1).into(),
-            None => panic!("no hostname found for ip address"),
+            None => panic!("no ip address found for a hostname"),
```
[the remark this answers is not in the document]
th7nder: Will do.

src/dns.rs
```diff
@@ -103,9 +109,18 @@ impl ToSocketAddrs for (String, u16) {
 
 impl<'a> ToSocketAddrs for (&'a str, u16) {
     fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
+        // When "x.y.a.b" is passed as a hostname.
```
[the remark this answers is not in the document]
th7nder: Will do.

src/dns.rs
```diff
@@ -1,8 +1,11 @@
 use indexmap::IndexMap;
 #[cfg(feature = "regex")]
 use regex::Regex;
-use std::net::{IpAddr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr, Ipv6Addr};
 
+/// Each new registered host has an IP in a subnet 192.168.0.0/24
+/// This is just a choice.
+/// localhost always resolves to 127.0.0.1, there is no localhost implementation for IPv6.
```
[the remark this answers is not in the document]
th7nder: Will do.

src/top.rs
```diff
@@ -358,10 +396,24 @@ impl Link {
                         dst: sent.dst,
                         message: sent.protocol,
                     };
+
+                    // If the communication is localhost, use a real ip address
+                    // of the host, instead of a localhost address to deliver a
+                    // message. One of them is guaranteed not to be a loopback
+                    // address.
+                    let ip_addr = if sent.src.ip().is_loopback() {
```
th7nder, line 404: We could try to workaround this and make communication work differently, however I don't know whether this is a good idea.
Currently localhost communication depends on e.g `192.168.0.1 <-> 127.0.0.1`. If we made it e.g. `127.0.0.1 <-> 127.0.0.1` then topology wouldn't be able to determine to which link it belongs.

One idea would be to make world `send_message`, host-aware, but not sure about that.

src/dns.rs
```diff
@@ -101,11 +103,17 @@ impl ToSocketAddrs for (String, u16) {
     }
 }
 
+
 impl<'a> ToSocketAddrs for (&'a str, u16) {
     fn to_socket_addr(&self, dns: &Dns) -> SocketAddr {
+        // When IP address is passed directly as a hostname.
```
mcches, line 109:
```suggestion
        // When IP address is passed directly as a str.
```

src/net/tcp/stream.rs
```diff
@@ -1,7 +1,7 @@
 use std::{
     fmt::Debug,
     io::{self, Result},
-    net::SocketAddr,
+    net::{SocketAddr},
```
mcches, line 4:
```suggestion
    net::SocketAddr,
```

src/top.rs
```diff
@@ -28,16 +28,38 @@ pub(crate) struct Topology {
 /// which orders the addrs, such that this type uniquely identifies the link
 /// between two hosts on the network.
```
mcches, line 29: Do you want to add a bit on the sort order with loopback now in play?

src/top.rs
```diff
@@ -130,6 +150,14 @@ impl<'a> Iterator for LinkIter<'a> {
     }
 }
 
+/// Network interface kind.
+/// Represents a different semantics of localhost and host-to-host communication.
```
mcches, line 154:
```suggestion
/// Differentiates the semantics of localhost and host-to-host communication.
```

src/top.rs
```diff
@@ -130,6 +150,14 @@ impl<'a> Iterator for LinkIter<'a> {
...
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+enum Interface {
+    Localhost,
```
mcches, line 157:
```suggestion
    Loopback,
```

src/top.rs
```diff
@@ -252,6 +280,16 @@ impl Topology {
...
+
+    fn validate(a: IpAddr, b: IpAddr) -> Pair {
+        let pair = Pair::new(a, b);
+        match pair.interface {
+            Interface::Localhost => {
+                panic!("can't perform topology operations on localhost interfaces")
+            }
+            Interface::External => pair,
+        }
+    }
```
mcches, line 292:
```suggestion
        if let Interface::Localhost = Pair::new(a, b).interface {
            panic!("can't perform topology operations on loopback")
        }
    }
```

src/world.rs
```diff
@@ -101,6 +101,12 @@ impl World {
 
         tracing::info!(target: TRACING_TARGET, hostname = ?self.dns.reverse(addr), ?addr, "New");
 
+        // Handles connections within a host
+        self.topology.register(addr, addr);
```
mcches, line 105: What's this one for?

tests/tcp.rs
```diff
@@ -23,7 +23,99 @@ fn assert_error_kind<T>(res: io::Result<T>, kind: io::ErrorKind) {
...
+
+    sim.client("server", async {
+        let listener = bind().await?;
+        tokio::spawn(async move {
+            loop {
+                let _ = listener.accept().await;
+            }
+        });
+
+        TcpStream::connect(("server", PORT)).await?;
```
mcches, line 41: I see, so you did (addr, addr) to make this work. Do we need it?

tests/tcp.rs
```diff
@@ -23,7 +23,99 @@ fn assert_error_kind<T>(res: io::Result<T>, kind: io::ErrorKind) {
...
+        assert_error_kind(
+            TcpStream::connect((Ipv4Addr::LOCALHOST, PORT)).await,
+            io::ErrorKind::ConnectionRefused,
+        );
+
+        Ok(())
+    });
+
+    sim.run()
 }
```
mcches, line 119: I would add another test that does at least one send and recv in each direction to test that logic. The tcp connection sequence is only in one direction since we cheat the syn-ack using a oneshot channel.

src/top.rs
mcches: This seems reasonable to me.

mcches (commented):
This is looking solid. I think we can merge after these comments are addressed.

tests/tcp.rs
th7nder: Actually, I don't think we don't. I'll remove it. If we need it in the future we can always add it.

src/world.rs
th7nder: Actually, I don't think we don't. I'll remove it. If we need it in the future we can always add it.

src/top.rs
th7nder: Will do.

tests/tcp.rs
```diff
@@ -23,7 +23,103 @@ fn assert_error_kind<T>(res: io::Result<T>, kind: io::ErrorKind) {
...
+    sim.run()
+}
+
+#[test]
+fn sends_data_within_a_localhost() -> Result {
+    let mut sim = Builder::new().build();
+
+    sim.client("server", async {
+        let listener = TcpListener::bind((Ipv4Addr::LOCALHOST, PORT)).await?;
+        tokio::spawn(async move {
```
mcches, line 79: You could avoid the spawn and unwrap to simplify this.

tests/tcp.rs
```diff
@@ -23,7 +23,103 @@ fn assert_error_kind<T>(res: io::Result<T>, kind: io::ErrorKind) {
...
+    sim.client("client", async {
+        assert_error_kind(
+            TcpStream::connect(("server", PORT)).await,
+            io::ErrorKind::ConnectionRefused,
+        );
+
+        assert_error_kind(
+            TcpStream::connect((Ipv4Addr::LOCALHOST, PORT)).await,
+            io::ErrorKind::ConnectionRefused,
+        );
```
mcches, line 117: I'd avoid this case. It has nothing to do with the server's ip.

mcches (approved):
Lgtm! A few minor comments.

tests/tcp.rs
th7nder: I think it does, it makes sure if we bind something to `localhost` it's not available publicly.

mcches: This is from a different host. How would it know which one to connect to even if localhost was exposed?

PR #84, opened 2023-02-13
Prevent JoinError::Cancelled to be propagated to the test on host crash

tereshch-aws:
When we crash host, we are dropping host's runtime causing host's task
to be cancelled. This change handles JoinError::Cancelled error for
finished tasks and ignores it.

### DIFF
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -1,4 +1,6 @@
-use crate::{for_pairs, Config, LinksIter, Result, Role, Rt, ToIpAddr, ToIpAddrs, World};
+use crate::{
+    for_pairs, Config, LinksIter, Result, Role, Rt, ToIpAddr, ToIpAddrs, World, TRACING_TARGET,
+};
 
 use indexmap::IndexMap;
 use std::cell::RefCell;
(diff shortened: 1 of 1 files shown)


src/sim.rs
```diff
@@ -299,15 +301,24 @@ impl<'a> Sim<'a> {
...
                     Role::Client { rt, handle } => (rt, handle),
                     Role::Simulated { rt, handle, .. } => (rt, handle),
                 };
-                rt.block_on(handle)??;
+                let result = rt.block_on(handle);
+                // Prevent JoinError::Cancelled to be propagated, assuming the task
+                // was crashed explicitly.
+                if let Err(e) = &result {
+                    if e.is_cancelled() {
+                        tracing::trace!(target: TRACING_TARGET, host = ?addr, "Crashed");
```
mcches, line 317: Should we trace right when the crash happens (in bounce or crash) rather than when we await the handle?

src/sim.rs
```diff
@@ -547,6 +558,20 @@ mod test {
         Ok(())
     }
 
+    /// This is a regression test that ensures JoinError::Cancelled is not
+    /// getting propagated to the test when the host crashes.
+    /// Before the fix the error returned by run function causing test to fail.
```
mcches, line 563: edit: ...is not propagated to the test when the host crashes, which was causing incorrect test failure.

src/sim.rs
tereshch-aws: I was looking into that but confused how to log it. Let me try to get it right.

PR #85, opened 2023-02-13
Fix crash causing join error cancelled propogation

tereshch-aws:


mcches (approved):
Made a few changes inline. Thanks!

PR #87, opened 2023-02-15
Set message latency per link

mcches:
We've got support for overriding the max, but it is also useful to
set a single latency duration for links.

Add another method on `Sim` to support this use case.


LucioFranco (approved):

tereshch-aws (approved):
lgtm

Issue #88, opened 2023-02-17
Re-starting a crashed host with bounce panics


mcches:
Repro:

```rust
#[test]
fn restart_host_after_crash() -> Result {
    let mut sim = Builder::new().build();

    sim.host("h", || async { future::pending().await });

    // crash and step to execute the err handling logic
    sim.crash("h");
    sim.step()?;

    // restart and step to ensure the host sfotware runs
    sim.bounce("h");
    sim.step()?;

    Ok(())
}
```

```txt
running 1 test
thread 'sim::test::restart_host_after_crash' panicked at 'missing host', src/sim.rs:143:43
stack backtrace:
   0: rust_begin_unwind
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:65:14
   2: core::panicking::panic_display
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:139:5
   3: core::panicking::panic_str
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:123:5
   4: core::option::expect_failed
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/option.rs:1879:5
   5: core::option::Option<T>::expect
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/option.rs:741:21
   6: turmoil::sim::Sim::run_with_hosts
             at ./src/sim.rs:143:22
   7: turmoil::sim::Sim::bounce
             at ./src/sim.rs:125:9
   8: turmoil::sim::test::restart_host_after_crash
             at ./src/sim.rs:606:9
   9: turmoil::sim::test::restart_host_after_crash::{{closure}}
             at ./src/sim.rs:596:5
  10: core::ops::function::FnOnce::call_once
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ops/function.rs:251:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ops/function.rs:251:5
```

On crash, the rt is removed causing this issue.
https://github.com/tokio-rs/turmoil/blob/main/src/sim.rs#L321-L322

PR #89, opened 2023-02-18
Handles host bounce after crash

tereshch-aws:
Fixing issue that host was removed from the sim state after host crash. With this fix, we change the way how we track state of the host during simulation. Crashed and hosts that have finished it's execution are no longer removed from simulation's state, that allowed us to bounce them. We might need to add explicit way to remove hosts from simulation in the future but there is no need to do it right now.

This change also encapsulates host's state management and it's operations(e.g. crash/bounce) into Role type that allowed to simplify Sim::step function code and make task's state management code more clean.

Update: @th7nder  submitted https://github.com/tokio-rs/turmoil/pull/89 in parallel, with the same approach to the fix but without refactoring. I don't mind to go with it if my refactoring is too much.

**Special focus:**
* Fixes https://github.com/tokio-rs/turmoil/issues/88
* Role enum -> struct refactoring. Role could be renamed now but I can't come up good name.
* I'm not happy `finished_err` function. It feels weird to check `is_finished` and then call `finished_err()`? but I was not able to come up with better way to do it.
* My initial idea to track hosts state as `Index<addr, Option<Role>>` didn't work out well. It polluted code with handling option is none and expects. After moving state into Role code become more clean.


mcches:
I agree `Role` has probably run its course. Maybe that should be called `Host` and the current one becomes `Os` since it is just per-host software that is always there.

src/role.rs
```diff
@@ -35,28 +45,74 @@ impl<'a> Role<'a> {
...
-        let rt = match self {
-            Role::Client { rt, .. } => rt,
-            Role::Simulated { rt, .. } => rt,
-        };
+    pub(crate) fn is_client(&self) -> bool {
+        matches!(self.role, RoleType::Client)
+    }
 
-        rt.now()
+    /// Runtime's current time
```
mcches, line 59: This...

src/role.rs
```diff
@@ -35,28 +45,74 @@ impl<'a> Role<'a> {
...
 
     pub(crate) fn tick(&self, duration: Duration) {
-        let rt = match self {
-            Role::Client { rt, .. } => rt,
-            Role::Simulated { rt, .. } => rt,
-        };
+        self.rt.tick(duration);
+    }
+
+    /// Returns true if task was finished
```
mcches, line 68: ...and this seems pretty self explanatory.

src/role.rs
```diff
@@ -35,28 +45,74 @@ impl<'a> Role<'a> {
...
+                Ok(res) => res,
+            }
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Crashes the host. Nothing will be running on the host after this call.
+    /// You can use [`Role::bounce`] to start the hosts up again.
+    pub(crate) fn crash(&mut self) {
```
mcches, line 95: Moving these here was a nice idea 👍

src/sim.rs
```diff
@@ -259,22 +250,26 @@ impl<'a> Sim<'a> {
...
     /// Runs each host in the simulation a fixed duration configured by
-    /// `tick_duration` in the builder.
+    /// `tick_duration` in the builder. Ignores hosts that already finished.
     ///
     /// The simulated network also steps, processing in flight messages, and
     /// delivering them to their destination if appropriate.
     pub fn step(&mut self) -> Result<bool> {
         let tick = self.config.tick;
 
+        // We stop simulation as soon as all clients finish
```
mcches, line 260: Let's move this up to the method docs.

```rust
///
/// Returns whether or not all clients have completed.
```

src/sim.rs
```diff
@@ -259,22 +250,26 @@ impl<'a> Sim<'a> {
     /// Step the simulation.
     ///
     /// Runs each host in the simulation a fixed duration configured by
-    /// `tick_duration` in the builder.
+    /// `tick_duration` in the builder. Ignores hosts that already finished.
```
mcches, line 253: I'm not sure this helpful to the caller.

src/sim.rs
```diff
@@ -259,22 +250,26 @@ impl<'a> Sim<'a> {
...
         let mut is_finished = true;
-        let mut finished = vec![];
 
         // Tick the networking, processing messages. This is done before
         // ticking any other runtime, as they might be waiting on network
         // IO. (It also might be waiting on something else, such as time.)
         self.world.borrow_mut().topology.tick_by(tick);
 
-        for (&addr, rt) in self.rts.iter() {
+        // Run each host, that is still running, in simulation and update it's
```
mcches, line 268: Consider the following:

```rust
// Tick each host with running software. If the software completes, 
// extract the result and return early if an error is encountered.
```

src/sim.rs
```diff
@@ -591,6 +561,36 @@ mod test {
         sim.run()
     }
 
+    /// This is a regression test that ensures that host could be restarted
```
mcches, line 564: The test name says this already.

src/role.rs
```diff
@@ -35,28 +45,74 @@ impl<'a> Role<'a> {
...
+    pub(crate) fn is_finished(&self) -> bool {
+        self.handle.as_ref().map_or(true, JoinHandle::is_finished)
+    }
 
-        rt.tick(duration)
+    /// Checks finished runtime for errors.
+    ///
+    /// Consumes the handle to extract task's execution result.
+    /// No-op if errors were already pulled or still task is still running
+    pub(crate) fn finished_err(&mut self) -> Result<()> {
```
mcches, line 77: Discussed a bit offline... perhaps something like:

```rust
fn is_finished(&mut self) -> Result<bool>
```

mcches (changes requested):

src/role.rs
tereshch-aws: This idea didn't work out well because we need to filter host with running software for loop in the `step` function. I've moved logic to extract error and drop handle into `Role::tick`.

tereshch-aws:
> I agree `Role` has probably run its course. Maybe that should be called `Host` and the current one becomes `Os` since it is just per-host software that is always there.

I've stuck with renaming. I like current `Host` name and it seems appropriate. I'm not huge fun of `OS` name. Host seems like combination of hardware and OS. 

Current `Role` seems like represent a software we run on the `Host`. I've tried a few options, I'm not super happy with all of them, first one is my favorite. 
* Role -> Application
* Role -> Service
* Role -> HostRuntime
* Role -> Soft

`RoleType` could be renamed to `Kind`. `Kind::Client` or `Kind::Server` or `Kind::Service`

mcches (commented):
Nice refactor idea for tick. Comments are mostly around naming and docs.

src/role.rs
```diff
@@ -11,21 +11,31 @@ use crate::{rt::Rt, Result};
 type Software<'a> = Box<dyn Fn() -> Pin<Box<dyn Future<Output = Result>>> + 'a>;
 
 /// Differentiates runtime fields for different host types
-pub(crate) enum Role<'a> {
+enum RoleType<'a> {
```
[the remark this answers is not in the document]
tereshch-aws: I think it is a great idea. Let me try to do it and see if there are any issues while trying. I'm not sure I will have time today though.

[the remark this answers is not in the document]
tereshch-aws: I think we could rename `RoleType` -> `SoftwareType` while moving, keeping Client, Simulated variants. Thoughts?

Optionally we can rename `Simulated`->`Server/Service` but I'm not sure it is a good ideal.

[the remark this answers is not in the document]
mcches: Sure, and we can do Client and Host for variants to match the public methods.

src/role.rs
```diff
@@ -35,28 +45,78 @@ impl<'a> Role<'a> {
...
+            None => Ok(true),
+        }
     }
 
-    pub(crate) fn tick(&self, duration: Duration) {
-        let rt = match self {
-            Role::Client { rt, .. } => rt,
-            Role::Simulated { rt, .. } => rt,
-        };
+    pub(crate) fn is_finished(&self) -> bool {
```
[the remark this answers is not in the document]
tereshch-aws: missed this comment. removed this method as separate commit

src/rt.rs
```diff
@@ -56,14 +137,54 @@ impl Rt {
...
+        } else if self.handle.take().is_some() {
+            self.cancel_tasks();
+        };
+    }
+
+    pub(crate) fn bounce(&mut self) {
+        if !self.is_host() {
+            panic!("can only bounce host's software");
+        } else {
+            self.cancel_tasks();
```
tereshch-aws, line 180: this code looks weird because cancel_tasks required mutable reference and borrow check was yelling at when I tried to use it inside of match block.

src/top.rs
```diff
@@ -21,7 +21,7 @@ pub(crate) struct Topology {
     /// We don't use a Rt for async. Right now, we just use it to tick time
     /// forward in the same way we do it elsewhere. We'd like to represent
     /// network state with async in the future.
-    rt: Rt,
+    rt: Rt<'static>,
```
tereshch-aws, line 24: This use of Rt was surprise to me and forced me to add Kind::NoSoftware but Kind::Client will work but I've thought adding explicit Kind is cleaner. `<'static>` hack is also awkward I had no idea it will work out. 

Using Rt for time only look strange to me in current state but it seems like we plan to extend it in future. Let me know if we need to change it.

src/rt.rs
```diff
@@ -1,43 +1,124 @@
...
+
+/// Type of runtime
+enum Kind<'a> {
+    /// A runtime that is running client software
+    Client,
+
+    /// A runtime that is running simulated host software
+    Host { software: Software<'a> },
+
+    /// A runtime that with no software running
```
mcches, line 24:
```suggestion
    /// A runtime without any software. The network topology uses this for 
    /// time tracking and message delivery.
```

src/rt.rs
```diff
@@ -1,43 +1,124 @@
...
 
-/// Per host simulated runtime
+// To support re-creation, we need to store a factory of the future that
+// represents the software. This is somewhat annoying in that it requires
+// boxxing to avoid generics.
+type Software<'a> = Box<dyn Fn() -> Pin<Box<dyn Future<Output = Result>>> + 'a>;
+
+/// Type of runtime
+enum Kind<'a> {
+    /// A runtime that is running client software
```
mcches, line 18:
```suggestion
    /// A runtime for executing test code.
```

src/rt.rs
```diff
@@ -1,43 +1,124 @@
...
+    Client,
+
+    /// A runtime that is running simulated host software
+    Host { software: Software<'a> },
+
+    /// A runtime that with no software running
+    NoSoftware,
+}
+
+/// Per host simulated runtime that runs software
```
mcches, line 28:
```suggestion
/// Per host simulated runtime.
```

src/rt.rs
```diff
@@ -1,43 +1,124 @@
...
     /// Handle to the Tokio runtime driving this simulated host. Each runtime
     /// may have a different sense of "now" which simulates clock skew.
     tokio: Runtime,
 
     /// Local task set used for running !Send tasks.
     local: LocalSet,
+
+    /// Software's handle
+    /// The handle is consumed as soon as software finishes or host crashes,
+    /// replaced with None to indicate finished software.
```
mcches, line 45:
```suggestion
    /// Optional handle to a host's software. When software finishes, the handle is
    /// consumed to check for error, which is propagated up to fail the simulation.
```

src/rt.rs
```diff
@@ -1,43 +1,124 @@
...
+        Self {
+            kind: Kind::Host { software },
+            tokio,
+            local,
+            handle: Some(handle),
+        }
+    }
+
+    // Creates special instance of `Rt` that has no running software and used
+    // only by Topology to advance simulated time.
```
mcches, line 86:
```suggestion
```

I'd suggest either documenting all of the constructors or none. The docs on `Kind` likely do the trick.

src/rt.rs
```diff
@@ -56,14 +137,54 @@ impl Rt {
     // 3. Other tasks on the `LocalSet` get a chance to run
     // 4. The sleep finishes
     // 5. The runtime pauses
```
mcches, line 139:
```suggestion
    // 5. The runtime pauses
    //
    // Returns whether the software has finished successfully or the error
    // that caused failure. Subsequent calls do not return the error as it is
    // expected to fail the simulation.
```

src/rt.rs
```diff
@@ -1,43 +1,124 @@
...
     }
 
+    /// Advances runtime's time forward by the specified `duration` and
+    /// allows software to run for give `duration.
+    ///
+    /// Returns whether the software has finished successfully or the error
+    /// that caused it to fail. The error could be returned only once, all
+    /// following calls will return Ok(true). Return Ok(false) if software is
+    /// still running.
+    //
```
mcches, line 121:
```suggestion
```

src/rt.rs
mcches: Let's merge and I have an idea for this that can be a follow up.

src/top.rs
mcches: `'static` seems appropriate to me here.

mcches (approved):
Nice changes. Comments are all doc suggestions.

PR #92, opened 2023-02-27
Add support for loopback in UDP

th7nder:
Completes  #79 , extends #83.

#83 introduced TCP support for loopback, as it is correct for IPv4, it was coincidentally working for IPv6 for some cases.
It was working in the test, but conceptually was wrong.  Something send from a localhost IPv6 interface, was treated in the code as sent from IPv4 interface.

If you bind to an unspecified interface, conceptually a message sent from it has to be sent from *some* interface. 
Each host has a IPv4 address generated, but it does not have an IPv6 address so besides `::1` there is no choice.
I decided sending from `::1` interface to non-loopback interfaces shouldn't be possible so I disabled the external communication for IPv6 via assertions. It'd look like `::1` is communicating with whatever external address in the traces.
 I leave IPv6 communication outside localhost as further issue to be investigated and implemented.

Another major change is with `World.send_message` requiring a host to be present.
Every communication has to go through the topology, but topology based solely `ip_addresses`, e.g. `127.0.0.1 <-> 127.0.0.1` cannot determine what host is the target. In the previous CR it was  worked around, by relying on the assumption that always one of the IP addresses won't be localhost. It wasn't possible in UDP so needed to change the approach.


th7nder:
Decided to go with a different approach, more details in #95

PR #93, opened 2023-02-28
Refactor TCP localhost communication

th7nder:
Related, #79, #83 

#83 introduced TCP support for loopback, as it is correct for IPv4, it was coincidentally working for IPv6 for some cases.
It was working in the test, but conceptually was wrong. Something send from a localhost IPv6 interface, was treated in the code as sent from IPv4 interface.

If you bind to an unspecified interface, conceptually a message sent from it has to be sent from some interface.
Each host has a IPv4 address generated, but it does not have an IPv6 address, so besides ::1 there is no choice.
I leave IPv6 communication  as further issue to be investigated and implemented.

Another major change is with World.send_message requiring a host to be present.
Every communication has to go through the topology, but topology based solely ip_addresses, e.g. 127.0.0.1 <-> 127.0.0.1 cannot determine what host is the target. In the previous CR it was worked around, by relying on the assumption that always one of the IP addresses won't be localhost.

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -1,10 +1,14 @@
-use std::{fmt::Display, net::SocketAddr};
+use std::{fmt::Display, net::{SocketAddr, IpAddr}};
 
 use bytes::Bytes;
 use tokio::sync::oneshot;
 
 #[derive(Debug)]
 pub(crate) struct Envelope {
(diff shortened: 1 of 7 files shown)


th7nder:
Decided to go with a different approach, more details in #95

PR #96, opened 2023-02-28
Add TCP loopback communication

th7nder:
#79

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -123,9 +123,18 @@ impl World {
     pub(crate) fn send_message(&mut self, src: SocketAddr, dst: SocketAddr, message: Protocol) {
         // Loopback communication does not use Topology.
         if self.sent_within_current_host(src, dst) {
-            let _ = self
-                .current_host_mut()
-                .receive_from_network(Envelope { src, dst, message });
+            if let Err(message) =
+                self.current_host_mut()
+                    .receive_from_network(Envelope { src, dst, message })
+            {
+                // Handle first response, ignore rest of them to avoid infinte loop.
+                // If it failed, it means RST. Can't do anything about it.
+                let _ = self.current_host_mut().receive_from_network(Envelope {
+                    src: dst,
+                    dst: src,
+                    message
+                });
+            }
         } else {
             self.topology
                 .enqueue_message(&mut self.rng, src, dst, message);
--- a/src/world.rs
+++ b/src/world.rs
@@ -1,4 +1,4 @@
-use crate::envelope::Protocol;
+use crate::envelope::{Envelope, Protocol};
 use crate::{config, Dns, Host, ToIpAddr, ToIpAddrs, Topology, TRACING_TARGET};
 
 use indexmap::IndexMap;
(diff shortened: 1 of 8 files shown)


src/world.rs
```diff
@@ -110,11 +110,26 @@ impl World {
...
+    }
+
+    /// Send `message` from `src` to `dst`.
+    /// Delivery between hosts is asynchronous and not guaranteed.
     pub(crate) fn send_message(&mut self, src: SocketAddr, dst: SocketAddr, message: Protocol) {
-        self.topology
-            .enqueue_message(&mut self.rng, src, dst, message);
+        // Loopback communication does not use Topology.
+        if self.sent_within_current_host(src, dst) {
+            let _ = self
```
th7nder, line 126: This is tricky. Should the communication on loopback be synchronous?
Topology has some queuing mechanism based on ticks, what about here?

mcches: Good question... I'd think about if two tasks could get into a busy loop chatting back and forth on the same host.

th7nder: In the proposed implementation, the tasks would communicate with each other as long they have something to send within a single step. They would block the current simulation step if they wanted to communicate. Only after they finished, simulation would proceed.

th7nder: I think introducing a queuing mechanism on a host level would solve the issue. Will try to implement that.

th7nder:
As we don't want to introduce queuing mechanism (offline talk with @mcches), we run into blocking issues with this PR.
Localhost CAN busy loop:
```rust
#[test]
fn localhost_communication_does_not_block_simulation() -> Result {
    if std::env::var("RUST_LOG").is_err() {
        std::env::set_var("RUST_LOG", "trace");
    }
    tracing_subscriber::fmt::init();
    let mut sim = Builder::new().build();

    sim.client("server", async {
        let listener = TcpListener::bind((Ipv4Addr::LOCALHOST, PORT)).await?;
        tokio::spawn(async move {
            if let Ok((mut stream, _)) = listener.accept().await {
                loop {
                    let _ = stream.write_u8(1).await;
                    assert_eq!(2, stream.read_u8().await.unwrap());
                }

            }
        });

        let mut stream = TcpStream::connect((Ipv4Addr::LOCALHOST, PORT)).await?;

        loop {
            assert_eq!(1, stream.read_u8().await?);
            stream.write_u8(2).await?;
        }
    });


    assert!(!sim.step()?);  
```

The issue is that writing on stream does not yield to
```rust
   pub(crate) fn tick(&mut self, duration: Duration) -> Result<bool> {
        self.tokio.block_on(async {
            self.local
                .run_until(async {
                    sleep(duration).await;
                })
                .await
        });
  ```
  
 A proposed solution is to implement a simpler channel for the Turmoil use case, example (courtesy of @mcches):
  ```rust
  #[test]
fn busy_bee() -> Result {
    let mut sim = Builder::new().build();

    sim.client("c1", async {
        let tx1 = Arc::new(Notify::new());
        let rx1 = tx1.clone();

        let tx2 = Arc::new(Notify::new());
        let rx2 = tx2.clone();

        tokio::spawn(async move {
            loop {
                rx1.notified().await;
                tx2.notify_waiters();
            }
        });

        loop {
            tx1.notify_waiters();
            rx2.notified().await;
        }
    });

    assert!(!sim.step()?);

    Ok(())
}
```
  
  
 However I don't have time to do that, so closing this PR and giving someone else a chance to implement it.
 Minor changes will come in separate PRs.

PR #101, opened 2023-03-10
feat: add try_* and readiness methods to UdpSocket

camshaft:
### Description

This change includes several tokio-compatible methods for UdpSocket:

* [try_send_to](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.try_send_to)
* [try_recv_from](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.try_recv_from)
* [readable](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.readable)
* [writable](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.writable)

I've also added the UdpSocket to the Send+Sync tests.

### Callouts

I needed to add a RX buffer since mpsc::Receiver doesn't have a way to just query if the channel has a message without taking the message.


src/net/udp.rs
```diff
@@ -1,26 +1,92 @@
...
 /// A simulated UDP socket.
 ///
 /// All methods must be called from a host within a Turmoil simulation.
 pub struct UdpSocket {
     local_addr: SocketAddr,
-    rx: Mutex<mpsc::Receiver<(Datagram, SocketAddr)>>,
+    rx: Mutex<Rx>,
+}
+
+struct Rx {
```
mcches, line 23: Nice solution! I'm going to take on a task to refactor this into the host similar to what TCP is doing. That should simplify some things.

mcches (approved):

PR #107, opened 2023-03-16
Fix AsyncRead impl for TcpStream

mcches:
Buffer the remainder of a received data segment to serve subsequent
reads.

### DIFF
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -173,16 +173,16 @@ impl ReadHalf {
     ///
     /// Returns an optional `Bytes` containing any remainder of `avail` that was
     /// not consumed.
-    fn put_slice(avail: Bytes, buf: &mut ReadBuf) -> Option<Bytes> {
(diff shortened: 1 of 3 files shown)


src/net/tcp/stream.rs
```diff
@@ -149,6 +167,24 @@ impl ReadHalf {
...
+        let amt = std::cmp::min(avail.len(), buf.remaining());
+        let (a, b) = avail.split_at(amt);
+
+        buf.put_slice(a);
+
+        if b.is_empty() {
+            None
+        } else {
+            Some(Bytes::copy_from_slice(b))
+        }
```
camshaft, line 186: Better to use `advance` so you don't have to copy the bytes out:

```suggestion
        if amt == 0 {
            return None;
        }
        
        buf.put_slice(&avail[..amt]);
        avail.advance(amt);

        if avail.is_empty() {
            None
        } else {
            Some(avail)
        }
```

camshaft (approved):

Issue #104, opened 2023-03-13
support ephemeral port assignments


camshaft:
I tried the following code:

```rust
#[test]
fn ephemeral_port() -> Result {
    let mut sim = Builder::new().build();

    sim.client("client", async {
        let sock = bind_to(0).await?;

        // turmoil should assign a port to the ephemeral range
        assert_ne!(sock.local_addr()?.port(), 0);
        assert!(sock.local_addr()?.port() >= 49152);

        Ok(())
    });

    sim.run()
}
```

It would be nice to support ephemeral port assignment. This is useful for clients that don't care about the specific port number; they just need a free port.

From https://www.rfc-editor.org/rfc/rfc6335#section-6:

>    o  the System Ports, also known as the Well Known Ports, from 0-1023
>      (assigned by IANA)
>
>   o  the User Ports, also known as the Registered Ports, from 1024-
>      49151 (assigned by IANA)
>
>   o  the Dynamic Ports, also known as the Private or Ephemeral Ports,
>      from 49152-65535 (never assigned)

th7nder:
Happy to help with this one, let's assign it to me.

PR #110, opened 2023-03-21
Support ephemeral port assignments on bind

th7nder:
#104

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -28,7 +28,7 @@ impl TcpListener {
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener. The port allocated can be queried via the `local_addr`
     /// method.
-    
+    ///
     /// Only 0.0.0.0 is currently supported.
     pub async fn bind<A: ToSocketAddrs>(addr: A) -> Result<TcpListener> {
         World::current(|world| {
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -26,6 +26,7 @@ impl TcpListener {
     /// The returned listener is ready for accepting connections.
     ///
     /// Only 0.0.0.0 is currently supported.
+    /// When the port is unspecified (0), an ephemeral port number is generated.
     pub async fn bind<A: ToSocketAddrs>(addr: A) -> Result<TcpListener> {
         World::current(|world| {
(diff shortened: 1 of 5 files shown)


src/net/udp.rs
```diff
@@ -94,6 +94,7 @@ impl UdpSocket {
     /// provided.
     ///
     /// Only 0.0.0.0 is currently supported.
+    /// When the port is unspecified (0), an ephemeral port number is generated.
```
mcches, line 97: Let's update this one to public docs as well:

```rust
/// This function will create a new UDP socket and attempt to bind it to
/// the `addr` provided.
///
/// Binding with a port number of 0 will request that the OS assigns a port
/// to this listener. The port allocated can be queried via the `local_addr`
/// method.
```

mcches (approved):
Thanks for doing this!

src/net/udp.rs
```diff
@@ -90,9 +90,12 @@ impl UdpSocket {
...
     }
 
-    /// Create a new simulated UDP socket and attempt to bind it to the `addr`
-    /// provided.
+    /// This function will create a new UDP socket and attempt to bind it to
+    /// the `addr` provided.
     ///
+    /// Binding with a port number of 0 will request that the OS assigns a port
+    /// to this listener. The port allocated can be queried via the `local_addr`
+    /// method.
```
mcches, line 98:
```suggestion
    /// method.
    
```

mcches (approved):

src/net/tcp/listener.rs
```diff
@@ -25,6 +25,9 @@ impl TcpListener {
     ///
     /// The returned listener is ready for accepting connections.
     ///
+    /// Binding with a port number of 0 will request that the OS assigns a port
+    /// to this listener. The port allocated can be queried via the `local_addr`
+    /// method.
```
mcches, line 30:
```suggestion
    /// method.
    
```

src/net/tcp/listener.rs
```diff
@@ -25,6 +25,10 @@ impl TcpListener {
...
+
```
mcches, line 31:
```suggestion
    ///
```

src/net/udp.rs
```diff
@@ -90,9 +90,13 @@ impl UdpSocket {
...
+
```
mcches, line 99:
```suggestion
    ///
```

PR #113, opened 2023-05-04
Add support for Ipv6 network/binds

PetrichorIT:
This PR implements a `IpNetwork` configuration option, to choose between a underlying Ipv4 or Ipv6 network.
The network parameter are used for:

a) granting addresses either from a `192.168.0.0/16` address space, or from `fe80:::/64`
b) binding sockets to associated unspecified addresses `0.0.0.0` / `:::`

The default configuration stays an Ipv4 network with `192.168.0.0/16`.


mcches:
Thanks for submitting the PR, and sorry for the delayed response. I'm curious to hear a bit more about why you need this feature for your simulation needs.

PetrichorIT:
No problem. We were trying to develop some units tests for Axum-based APIs in a system, that for the most part does only support Ipv6 addressing. Changing the addressing either to Ipv4Addr or to IpAddr would be a lot of work and may result in changes in the systems behavior. In short, the codebase is quite large, so we would like to use the least intrusive method to simulate it, which would be Ipv6 support on sockets.

src/builder.rs
```diff
@@ -9,6 +9,8 @@ pub struct Builder {
 
     config: Config,
 
+    network: IpNetwork,
```
mcches, line 12: How about the following?

```suggestion
    interface: IpInterface
```

mcches, line 12: How about the following?

```suggestion
    ip_version: IpVersion
```

src/config.rs
```diff
@@ -13,6 +16,66 @@ pub(crate) struct Config {
     pub(crate) epoch: SystemTime,
 }
 
+/// The kinds of networks that can be simulated in turmoil
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
```
mcches, line 20: Are all of these derives necessary?

src/config.rs
```diff
@@ -13,6 +16,66 @@ pub(crate) struct Config {
...
+pub enum IpNetwork {
```
mcches, line 21: Perhaps housing this in `dns.rs` or its own file `ip.rs` would be better.

src/dns.rs
```diff
@@ -1,11 +1,13 @@
 use indexmap::IndexMap;
 #[cfg(feature = "regex")]
 use regex::Regex;
-use std::net::{IpAddr, SocketAddr, Ipv4Addr, Ipv6Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+
+use crate::config::IpNetworkAddrIter;
 
 /// Each new host has an IP in the subnet 192.168.0.0/24.
```
mcches, line 8: Can you update this doc?

src/net/tcp/listener.rs
```diff
@@ -38,7 +38,11 @@ impl TcpListener {
             if !addr.ip().is_unspecified() {
                 panic!("{addr} is not supported");
             }
-            
+
+            if addr.is_ipv4() != host.addr.is_ipv4() {
+                panic!("{addr} is not supported (Ipv4/Ipv6 missmatch)")
```
mcches, line 43:
```suggestion
                panic!("ip version mismatch: {:?} host: {:?}", addr, host.addr)
```

src/net/udp.rs
```diff
@@ -106,7 +106,11 @@ impl UdpSocket {
             if !addr.ip().is_unspecified() {
                 panic!("{addr} is not supported");
             }
-            
+
+            if addr.is_ipv4() != host.addr.is_ipv4() {
+                panic!("{addr} is not supported (Ipv4/Ipv6 missmatch)")
```
mcches, line 111:
```suggestion
                panic!("ip version mismatch: {:?} host: {:?}", addr, host.addr)
```

tests/ipv6.rs
```diff
@@ -0,0 +1,81 @@
+use std::{net::IpAddr, str::FromStr};
```
mcches, line 1: Could you move the address specific test to the file where you define the ip version and iterator? Then the udp and tcp specifics can go in their respective test files.

src/config.rs
PetrichorIT: Hash may be unnessecar, and Copy is a matter of preference, but i think having 
PartialEq, Eq would be good since match!(x, y) is less intuitive than x == b

src/builder.rs
PetrichorIT: i think this one is good

src/ip.rs
```diff
@@ -0,0 +1,117 @@
...
+impl IpVersion {
+    pub(crate) fn iter(&self) -> IpVersionAddrIter {
+        match self {
+            Self::V4 => IpVersionAddrIter::V4(1),
+            Self::V6 => IpVersionAddrIter::V6(1),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
mcches, line 22: nit: I don't think you want to copy or clone (or probably any of these derives) this as you'd have different next return values across each instance.

mcches (approved):
Looks good to me.

PR #114, opened 2023-05-16
Remove inconsistent address resolution

PetrichorIT:
There is an inconsistency in address resolution / parsing.
The trait `ToIpAddr` interprets strings that can be parsed as an `IpAddr` as 
domain names instead of parsing the address. `ToSocketAddr` however will
parse the Ip part of the address if possible. To remove this inconsistency 
`ToIpAddr` will now always interpret strings that can be parse to a `IpAddr` as `IpAddr`.

# Example

Without this change:
```rust
use turmoil::{*, net::*};

fn main() -> Result {
    let mut sim = Builder::new().build();
    sim.client("192.168.42.42", async move {
        let addr = UdpSocket::bind("0.0.0.0:0").await.unwrap().local_addr().unwrap();
        // addr will be 192.168.0.1:<port>, interpreting the clients name as a domain addr
        Ok(())
    });
    sim.client("192.168.57.57".parse::<IpAddr>().unwrap(), async move {
        let addr = UdpSocket::bind("0.0.0.0:0").await.unwrap().local_addr().unwrap();
        // addr will be 192.168.57.57:<port>, respecting the provided parameter
        Ok(())
    });
   
   // The problem is even more pronounced with lookups
   sim.client("x", async move {
        // lookups will use the ip->ip dns entry
        assert_eq!(lookup("192.168.42.42"), Ipv4Addr::new(192,168,0,1));
       
       // ToSocketAddr however not use dns but rather parse the raw ip part
       // Thus this will panic, since the link `x->192.168.42.42`does not exits
       let stream = TcpStream::connect("192.168.42.42:80").await?;
       Ok(()) 
   });

  sim.run()
}
```

On a site note, `ToIpAddr` was not implemented for `Ipv4Addr` and `Ipv6Addr` which 
was problematic since APIs with `impl ToIpAddr` could not use `.into()` due to missing 
type inference. Same goes for `ToSocketAddr` with `SocketAddrV4` and `SocketAddrV6` respectivly
(Tokio's `ToSocketAddr` does is implemented for these types)

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,7 +1,7 @@
 use indexmap::IndexMap;
 #[cfg(feature = "regex")]
 use regex::Regex;
-use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 
 use crate::ip::IpVersionAddrIter;
 
@@ -66,6 +66,11 @@ impl ToIpAddr for String {
 
 impl<'a> ToIpAddr for &'a str {
     fn to_ip_addr(&self, dns: &mut Dns) -> IpAddr {
+        let ipaddr: Result<IpAddr, _> = self.parse();
+        if let Ok(ipaddr) = ipaddr {
+            return ipaddr;
+        }
+
         *dns.names
             .entry(self.to_string())
             .or_insert_with(|| dns.addrs.next())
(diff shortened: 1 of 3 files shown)


src/dns.rs
```diff
@@ -66,6 +66,11 @@ impl ToIpAddr for String {
 
 impl<'a> ToIpAddr for &'a str {
     fn to_ip_addr(&self, dns: &mut Dns) -> IpAddr {
+        let ipaddr: Result<IpAddr, _> = self.parse();
+        if let Ok(ipaddr) = ipaddr {
+            return ipaddr;
```
mcches, line 71:
```suggestion
        if let Ok(ip) = self.parse::<IpAddr> {
            return ip;
```

mcches (approved):
Thanks for making the Addrs more complete!

PR #117, opened 2023-05-19
Fix bug in binding an in use port

mh32:
There was a bug where binding an already bound port would wipe the existing bind before returning an ADDR_IN_USE error. This change resolves the bug by erroring before we update the host's binds.


mcches (approved):
Lgtm!

PR #118, opened 2023-05-19
Add support for loopback

mh32:
This changes adds UDP and TCP support for loopback. Packets sent over loopback
do not flow through the topology, instead they are sent directly from the host
back to itself - somewhat modeling how the real world works. The affects of
this implementation are that network manipulation APIs do not work on loopback
traffic. However, this approach gets incrementally closer towards full loopback
support and unblocks most use cases.

The implementation is a bit of hack. I think where we want to get to is full 
support for multiple interfaces. For example, hosts should be allowed to have 
both an v4 and v6 ip addr. This will require a bit of a lift due to how hosts 
are identified by an IP address.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -29,13 +29,13 @@ impl TcpListener {
     /// to this listener. The port allocated can be queried via the `local_addr`
     /// method.
     ///
-    /// Only `0.0.0.0` or `::` are currently supported.
+    /// Only `0.0.0.0`, `::`, or localhost are currently supported.
     pub async fn bind<A: ToSocketAddrs>(addr: A) -> Result<TcpListener> {
         World::current(|world| {
             let mut addr = addr.to_socket_addr(&world.dns);
             let host = world.current_host_mut();
 
-            if !addr.ip().is_unspecified() {
+            if !addr.ip().is_unspecified() && !addr.ip().is_loopback() {
                 panic!("{addr} is not supported");
             }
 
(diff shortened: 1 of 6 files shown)


src/net/tcp/listener.rs
```diff
@@ -29,22 +29,20 @@ impl TcpListener {
...
+            if !addr.ip().is_unspecified() && !addr.ip().is_loopback() {
                 panic!("{addr} is not supported");
             }
 
             if addr.is_ipv4() != host.addr.is_ipv4() {
                 panic!("ip version mismatch: {:?} host: {:?}", addr, host.addr)
             }
 
-            // Unspecified -> host's IP
-            addr.set_ip(host.addr);
```
mcches, line 47: Nice!

src/net/tcp/stream.rs
```diff
@@ -259,10 +269,34 @@ impl WriteHalf {
...
+                world
+                    .current_host_mut()
+                    .receive_from_network(Envelope { src, dst, message });
+            if let Err(rst) = result {
+                let _ = world.current_host_mut().receive_from_network(Envelope {
+                    src: dst,
+                    dst: src,
+                    message: rst,
+                });
+            }
```
mcches, line 295:
```suggestion
            if let Err(rst) = world
                .current_host_mut()
                .receive_from_network(Envelope { src, dst, message }) {
                _ = world.current_host_mut().receive_from_network(Envelope {
                    src: dst,
                    dst: src,
                    message: rst,
                });
            }
```

mcches (approved):
This looks great. Could you please update #79 with our thoughts on where this should go long term?

PR #120, opened 2023-05-31
Fix reverse dns lookups on static binds

PetrichorIT:
Turmoil supports the creation of clients/hosts bound to any address that `impl ToIpAddr`.
This allows for then creation of nodes, bound to static IP addresses. Such nodes
will not appear in the internal mappings of `Dns`, since they have no hostname.

`Dns::reverse` however expects that all nodes (all inputs would be more correct) are contained
in the dns mapping. If this assumption is broken, `Dns::reverse` panics. This means that any use of
`Dns::reverse` in combination with a statically bound address will panic. 

The problem: If a tracing subscriber is active, internal traces will attempt to use `Dns::reverse` once
when the node is initialized, thus crashing the simulation (world.rs:107).

## Solutions
I See three solutions:

### A (currently implemented)
If a reverse lookup does not find any entry in the dns mapping, it returns the input address as a `String`.
This would keep API changes small (`&str` to `String`). However all input addresses that have not appeared 
in the mapping would be handled this way, not just statically bound addresses.

### B
Add static binds to dns mapping. We could implement node creation in a way, that a dns entry
mapping the stringified  static address to the real static address is created. This would not
change the `Dns::reverse`, but major changes in address resolution / node creation would be nessecary.

### C
Change the API of `Dns::reverse` to reflect the failable nature of such lookups. By changing the return type to either 
`Option<&str>` or `Result<&str>` we could deal with failing lookups without panicking. This would be the most expressiv
option, but it would change the public API of `Dns::reverse` and `Sim::reverse_lookup_pair`.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -49,12 +49,11 @@ impl Dns {
         addrs.to_ip_addrs(self)
     }
 
-    pub(crate) fn reverse(&self, addr: IpAddr) -> String {
+    pub(crate) fn reverse(&self, addr: IpAddr) -> Option<&str> {
         self.names
             .iter()
             .find(|(_, a)| **a == addr)
-            .map(|(name, _)| name.to_string())
-            .unwrap_or_else(|| addr.to_string())
+            .map(|(name, _)| name.as_str())
     }
 }
 
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -49,12 +49,12 @@ impl Dns {
         addrs.to_ip_addrs(self)
     }
 
-    pub(crate) fn reverse(&self, addr: IpAddr) -> &str {
+    pub(crate) fn reverse(&self, addr: IpAddr) -> String {
         self.names
             .iter()
             .find(|(_, a)| **a == addr)
-            .map(|(name, _)| name)
-            .expect("no hostname found for ip address")
+            .map(|(name, _)| name.to_string())
+            .unwrap_or_else(|| addr.to_string())
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/dns.rs
```diff
@@ -49,12 +49,12 @@ impl Dns {
         addrs.to_ip_addrs(self)
     }
 
-    pub(crate) fn reverse(&self, addr: IpAddr) -> &str {
+    pub(crate) fn reverse(&self, addr: IpAddr) -> String {
```
mcches, line 52: Perhaps we can make the internal version fallible and leverage that within the tracing calls, but keep the public API as is. So this would return `Option<&str>` and then we just `.expect(..)` on it for now. We can consider making the public ones fallible too in a later release.

This was purely added for debugging convenience as its easier to look at `&str` hostnames instead of IP addresses.

mcches (commented):
Nice catch! Let me know what you think about the hybrid approach.

src/dns.rs
PetrichorIT: Seems like the best solution for now.

tests/dns.rs
```diff
@@ -0,0 +1,12 @@
+use std::net::Ipv4Addr;
+use turmoil::*;
+
+// This test is not in src/dns.rs since it initalizes a tracing subscriber
+// which will presist, and effect other tests in the same file.
```
mcches, line 5:
```suggestion
// which will persist, and effect other tests in the same file.
```

mcches (approved):
Lgtm! Thanks for all the recent improvements.

PR #121, opened 2023-06-05
Make socket buffer capacity configurable

mcches:
Behavior stays the same for now, but the values can at least be modified for chatty use cases.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -275,7 +275,7 @@ impl StreamSocket {
     // Buffer and re-order received segments by `seq` as the network may deliver
     // them out of order.
     fn buffer(&mut self, seq: u64, segment: SequencedSegment) -> Result<(), Protocol> {
-        use mpsc::error::TrySendError::Closed;
+        use mpsc::error::TrySendError::*;
 
         let exists = self.buf.insert(seq, segment);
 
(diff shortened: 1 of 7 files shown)


mcches:
We need high fidelity behavior (e.g. back-pressure), but this is a no-op behavior-wise and unblocks people so I think it is worth doing now.

tests/udp.rs
```diff
@@ -559,3 +559,38 @@ fn localhost_ping_pong() -> Result {
...
+            .is_err());
+
+        Ok(())
+    });
+
+    sim.client("client", async move {
+        let s = bind().await?;
+
+        send_ping(&s).await?;
+        send_ping(&s).await?; // dropped
```
paulyeo21, line 589: how does this test that the packet was dropped? this is suppose to hit L182 of host.rs, correct?

paulyeo21 (approved):
i had one clarifying question, otherwise LGTM

tests/udp.rs
mcches: We send 2 datagrams and then notify the peer to recv them. With a capacity of 1, the first is delivered and enqueued and the second is dropped. Line 578 verifies that the second was never delivered by timing out on recv.

paulyeo21: oh gotcha. thanks!

src/host.rs
```diff
@@ -288,7 +287,7 @@ impl StreamSocket {
             let segment = self.buf.remove(&self.recv_seq).unwrap();
             self.sender.try_send(segment).map_err(|e| match e {
                 Closed(_) => Protocol::Tcp(Segment::Rst),
-                _ => todo!("{} socket buffer full", self.local_addr),
+                _ => panic!("{} socket buffer full", self.local_addr),
```
tereshch-aws, line 290: Nit: maybe replace `_` with `Full` variant?

tereshch-aws (approved):

PR #124, opened 2023-06-20
Tracing improvments

PetrichorIT:
Some improvements to tracing with turmoil:
- All nodes now have a node-span, to add information about the currently active node to all tracing events
- These spans are stored in `Rt` and `Host` to cover all codepaths that may emit events
- All internal tracing events now share a common structure (`src`,`dst`, ..., `msg`)
- `src` now always represents the source of the packet, the tracing record was about  (`src` of SYN is client, `src` of SYN-ACK is server)

### DIFF
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -7,6 +7,7 @@ use std::net::IpAddr;
 use std::ops::DerefMut;
 use std::time::UNIX_EPOCH;
 use tokio::time::Duration;
+use tracing::Level;
 
 /// A handle for interacting with the simulation.
 pub struct Sim<'a> {
@@ -65,15 +66,23 @@ impl<'a> Sim<'a> {
         F: Future<Output = Result> + 'static,
     {
         let addr = self.lookup(addr);
+        let name = self
+            .world
+            .borrow_mut()
+            .dns
+            .reverse(addr)
+            .map(str::to_string)
+            .unwrap_or_else(|| addr.to_string());
+        let span = tracing::span!(Level::INFO, "node", name);
 
         {
             let world = RefCell::get_mut(&mut self.world);
 
             // Register host state with the world
-            world.register(addr, &self.config);
+            world.register(addr, span.clone(), &self.config);
         }
 
-        let rt = World::enter(&self.world, || Rt::client(client));
+        let rt = World::enter(&self.world, || Rt::client(span, client));
 
         self.rts.insert(addr, rt);
     }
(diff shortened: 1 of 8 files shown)


davidbarsky:
Generally speaking, you shouldn't need to pass a `tracing::Span` to usage sites: you can create a new one at the usage site and `tracing` will figure out the appropriate parentage/hierarchy automatically.

mcches:
Tagging #49.

PetrichorIT:
@davidbarsky since spans are entered 3 times pre tick, it seems wasteful to me to create the span inline,
since they require custom fields generated from a reverse dns request each time. The core idea behind passing
the span to the appropriate usage site is to minimize dns queries and string allocation. Would you consider the performance impact of span creation significant enough to warrant this unusual structure or should we just switch to inline-creation of spans, say with a shared `Arc<String>` to minimize reverse dns queries?

davidbarsky:
> Would you consider the performance impact of span creation significant enough to warrant this unusual structure or should we just switch to inline-creation of spans, say with a shared `Arc<String>` to minimize reverse dns queries?

Creating spans and events is generally pretty cheap, _especially_ if they are disabled and filtered out. I think keeping an `Arc<str>` around, especially if it's expensive to fetch, is perfectly reasonable.

src/top.rs
```diff
@@ -297,7 +297,7 @@ impl Link {
         dst: SocketAddr,
         message: Protocol,
     ) {
-        tracing::trace!(target: TRACING_TARGET, ?src, ?dst, protocol = %message, "Send");
+        tracing::trace!(target: TRACING_TARGET, ?src, ?dst, protocol = %message, "Send (2)");
```
mcches, line 300: Why "Send (2)" here?

src/sim.rs
```diff
@@ -98,7 +116,7 @@ impl<'a> Sim<'a> {
             world.register(addr, &self.config);
```
mcches, line 116: Can you update this event? https://github.com/tokio-rs/turmoil/blob/main/src/world.rs#L108 It could use the `nodename` instead of rendering the optional.

mcches (approved):
A couple minor comments. This will be really nice for debugging.

src/top.rs
PetrichorIT: Oh, that should have been removed, that was only there for debugging purposes.

src/sim.rs
PetrichorIT: like this?

PR #129, opened 2023-07-17
Add hold, release, repair, partition to simulator

tthebst:
Expose functions to control network behaviour on the simulator.

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -1,7 +1,7 @@
 use crate::config::Config;
 use crate::envelope::Protocol;
 use crate::ip::IpVersionAddrIter;
-use crate::{config, Dns, Host, ToIpAddr, ToIpAddrs, Topology, TRACING_TARGET};
+use crate::{config, for_pairs, Dns, Host, ToIpAddr, ToIpAddrs, Topology, TRACING_TARGET};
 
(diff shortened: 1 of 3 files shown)


mcches (commented):
Seems useful, but can you factor it to remove the duplication between these and the free functions that already exist?

src/world.rs
```diff
@@ -1,7 +1,7 @@
 use crate::config::Config;
 use crate::envelope::Protocol;
 use crate::ip::IpVersionAddrIter;
-use crate::{config, Dns, Host, ToIpAddr, ToIpAddrs, Topology, TRACING_TARGET};
+use crate::{config, for_pairs, Dns, Host, ToIpAddr, ToIpAddrs, Topology, TRACING_TARGET};
```
mcches, line 4: Let's also move `for_pairs` into here since I believe that is the only place we use it now.

mcches (approved):
One minor comment. Lgtm!

src/world.rs
tthebst: It is also used in `sim.rs` should it still be moved to `world.rs`?

mcches: Ok, pub(crate) on the lib seems fine.

Issue #127, opened 2023-07-14
Return errors instead of panicking, when sending invalid packets.


PetrichorIT:
Currently `turmoil` will panic, if a packet is send to an ip address that does not exist,
since this will result in an invalid access to the index map in `top.rs`.

This does not mirror the behavior if `tokio` or `std` sockets and panicking seems too extrem,
especially since some applications may create such sockets, expecting errors instead of
panics.

Therefore it might be advantageous to return errors instead of panicking in `World::send_message`.

# Example 

This example will panic.
```rust
fn main() -> Result {
     let mut sim = Builder::new().build();
     sim.client("client", async move {
         let _ = net::TcpStream::connect("192.168.30.1:80").await?;
         Ok(())
     });

     sim.run()
 }
```

```
thread 'main' panicked at 'IndexMap: key not found', ~/dev/turmoil/src/top.rs:221:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: JoinError::Panic(Id(1), ...)
```

mcches:
This seems like a good change to make. Do you want to take a stab at it?

PetrichorIT:
Yeah, I would like to do it.

PR #130, opened 2023-07-18
Return errors instead of panicking

PetrichorIT:
See Issue #127.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -42,15 +42,12 @@ impl TcpListener {
             if !addr.ip().is_unspecified() && !addr.ip().is_loopback() {
                 return Err(Error::new(
                     ErrorKind::AddrNotAvailable,
-                    "invalid argument - addr not supported",
+                    format!("{addr} is not supported"),
                 ));
             }
 
             if addr.is_ipv4() != host.addr.is_ipv4() {
-                return Err(Error::new(
-                    ErrorKind::InvalidInput,
-                    "invalid argument - ip version missmatch",
-                ));
+                panic!("ip version mismatch: {:?} host: {:?}", addr, host.addr)
             }
 
             if addr.port() == 0 {
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -1,4 +1,8 @@
-use std::{io::Result, net::SocketAddr, sync::Arc};
+use std::{
+    io::{Error, ErrorKind, Result},
+    net::SocketAddr,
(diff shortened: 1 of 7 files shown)


PetrichorIT:
On a related note: 
It may be beneficial to apply the error-over-panic mentality to more cases, notably `TcpListener::bind` and `UdpSocket::bind`.  Should we do this within this PR ?

mcches:
> On a related note: It may be beneficial to apply the error-over-panic mentality to more cases, notably `TcpListener::bind` and `UdpSocket::bind`. Should we do this within this PR ?

Yes, that makes sense to me.

src/net/tcp/listener.rs
```diff
@@ -36,11 +40,17 @@ impl TcpListener {
             let host = world.current_host_mut();
 
             if !addr.ip().is_unspecified() && !addr.ip().is_loopback() {
-                panic!("{addr} is not supported");
```
mcches, line 39: Can you keep the same message format for these? Having the arguments in the error message make it faster to debug the issue.

src/net/tcp/listener.rs
```diff
@@ -36,11 +40,17 @@ impl TcpListener {
...
+                return Err(Error::new(
+                    ErrorKind::AddrNotAvailable,
+                    "invalid argument - addr not supported",
+                ));
             }
 
             if addr.is_ipv4() != host.addr.is_ipv4() {
```
mcches, line 49: When we do the larger refactor this will be handled, but for now this seems like a panic. You'd never be handling this in your application, it would be a misconfiguration of the sim.

src/top.rs
```diff
@@ -216,9 +217,13 @@ impl Topology {
...
         message: Protocol,
-    ) {
-        let link = &mut self.links[&Pair::new(src.ip(), dst.ip())];
-        link.enqueue_message(&self.config, rand, src, dst, message);
+    ) -> Result<()> {
+        if let Some(link) = self.links.get_mut(&Pair::new(src.ip(), dst.ip())) {
+            link.enqueue_message(&self.config, rand, src, dst, message);
+            Ok(())
+        } else {
+            Err(Error::new(ErrorKind::ConnectionRefused, "host unreachable"))
```
mcches, line 225: nit: The error message renders like this:

```
Err(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
```

mcches (approved):
This looks good, but I'm still thinking ip version mismatch is a panic with the current setup.

PR #133, opened 2023-07-26
Add message duplication for udp

tthebst:
Adds ability to configure message duplication.

Some questions: 
1. This should only apply to UDP right? It does not make sense for TCP.
2. Should I name it `udp_message_duplication` instead?
3. I was a little unsure where to put the duplication. In the end I thought it makes sense to put in `enqueue`. Do you think this is the right place?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,6 +27,9 @@ pub(crate) struct Link {
 
     /// How often sending a message works vs. the message getting dropped
     pub(crate) message_loss: Option<MessageLoss>,
+
+    /// How often a sent message is duplicated. Only applies to UDP.
+    pub(crate) message_duplication: Option<MessageDupliaction>,
 }
 
 /// Configure latency behavior between two hosts.
@@ -50,6 +53,13 @@ pub(crate) struct MessageLoss {
 
     /// Probability of a failed link returning
     pub(crate) repair_rate: f64,
+}
+
+/// Configure how often messages are duplicated
+#[derive(Clone)]
+pub(crate) struct MessageDupliaction {
+    /// Probability of a link duplicating messages. Only applies to UDP.
+    pub(crate) duplication_rate: f64,
 }
 
 impl Default for Config {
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -27,6 +27,9 @@ pub(crate) struct Link {
 
     /// How often sending a message works vs. the message getting dropped
     pub(crate) message_loss: Option<MessageLoss>,
+
+    /// How often a sent message is duplicated. Only applies to UDP.
+    pub(crate) message_duplication: Option<MessageDupliaction>,
```
mcches, line 32: Do we need a struct? can this just be a `f64` with `0.0` for a default value?

mcches:
> Should I name it udp_message_duplication instead?

I think `duplication_rate` is fine. The rustdoc on the pub part can explain that it is for UDP only.

src/top.rs
```diff
@@ -334,6 +345,21 @@ impl Link {
             }
         };
 
+        match message {
+            // Duplicate UDP messages
+            Protocol::Udp(ref datagram)
+                if self.rand_duplication(global_config.message_duplication(), rand) =>
+            {
+                self.sent.push_back(Sent {
```
mcches, line 353: I'd like to get @camshaft's take here on what would be useful.

mcches (changes requested):

src/config.rs
levinotik, line 32: Typo here. Should be `duplication` not `dupliaction`

camshaft: The struct name also has the typo

src/top.rs
camshaft: In s2n-quic, [we duplicate before we apply any delay or corruption](https://github.com/aws/s2n-quic/blob/f7dd7844486141acdf4ecbf68feb752c042d191f/quic/s2n-quic-platform/src/io/testing/model.rs#L323-L334) in order to have each packet be treated separately. This enables the simulation to interleave duplicate packets with others rather than having a burst of duplicate packets since [each packet gets its own delivery time](https://github.com/aws/s2n-quic/blob/f7dd7844486141acdf4ecbf68feb752c042d191f/quic/s2n-quic-platform/src/io/testing/model.rs#L309). We also loop up to 5 times instead of a single duplication, may be useful depending on the application under test.

tthebst: Do you think it makes sense to make this configurable with a single float? For example if you specify duplication 5.0 it duplicates five times and 5.5 would be between five and six.

camshaft: it depends on what the intent is... with the s2n-quic implementation each iteration is exponentially less likely to happen since the loop bails on the first retransmission fails:

```rust
            let mut count = 0;
            while count < 5 && gen_rate(retransmit_rate) {
                transmission_count += transmit(Cow::Borrowed(&packet));
                count += 1;
            }

            transmission_count += transmit(Cow::Owned(packet));
```

If you have a single float, it wouldn't be the same thing. Not saying it needs to be the same. But I'm not sure how useful it is to always duplicate packets in fixed bursts like that. IMO the exponential backoff is closer to how real networks behave.

mcches:
Closing due to inactivity. Feel free to re-open.

Issue #132, opened 2023-07-25
Refactoring: Adding mixed IP subnets


PetrichorIT:
This refactor aims to introduce the ability of nodes to have multiple addresses
in distinct subnets.

# Immediate Goals

- Each node should be bound to an unique `Ipv4Addr` AND an unqiue `Ipv6Addr`
- All bound addresses should be in a predefined subnet (like `192.168.0.0/16`)
- The available subnets should be configurable in the `Builder`
- Addresses can be either automatically or manually assigned

# Challenges

I have tried to implement this, and come to the conlsuion that some major changes
internally AND externally would be nessecary. Notably:

- Nodes, and thus `Rt`/`Host` can no longer by identified by a single `IpAddr`. 
  The best possible solution would be to identify them by something like a MAC addr,
  but that would warrent major internal changes
- `lookup` would need to return more than one possible address, thus the public API
  of `ToIpAddr` / `lookup` / `lookup_many` would need to change. This could be a good
  moment to introduce API compliance with either `std::net` or `tokio::net`
- using `ToIpAddr` for module creation creates problems when statically assigning addresses.
  Even without this refactoring, nodes with explicitly assigned address cannot have human readable
  names, since their place is traded for the address assignment. Mixed IP subnets only enlarge this
  problem. The current api of `Sim::client` / `Sim::host` provides no way to explicitly assign both
  an Ipv4 and an Ipv6 address. In short, the current API cannot support a node with explicitly assigned 
  addresses, let alone a human-readable name, so changes would be nessecary.

In my opinion, this amount of changes would exceed the scope of one PR, so it might be benifical
to make step by step changes to the public API. However this warrants discussion.

# Some related thoughts

- While not in the scope of this refactoring, binding sockets to specific addresses may be beneficial
- in `std`/`tokio` Ipv6 sockets bound to `[::]` can receive incoming Ipv4 packets (addresses are being mapped to Ipv6),
  however the reverse is not possible. This seems like an rare edge case, so i do not know whether we 
  should ever support this behaviour
- I might prove useful in the future, to refrain from hardcoding only two possible addresses in two possible subnets per node.
  Supporting a set of bound addresses+subnets might be beneficial to a) remodel localhost, to use `top.rs`links or b) add support for multiple interface, thus multiple subnets, should that ever be a goal

As a reference, my current test implementation can be found [here](https://github.com/PetrichorIT/turmoil/tree/ipnetworks).
[...]

PR #140, opened 2023-09-14
Extend loopback workaround

### LINKED ISSUE #135: Loopback is incomplete
The following scenarios exist for client -> server within the same host:

(Only Tcp is shown, but we need to handle it for Udp as well)
```
// bind | connect

// 0s | 127.0.0.1
// client: local Ok(127.0.0.1:49582), peer Ok(127.0.0.1:1234)
// server: 127.0.0.1:49582, local Ok(127.0.0.1:1234), peer Ok(127.0.0.1:49582)

// 127.0.0.1 | 127.0.0.1
// client: local Ok(127.0.0.1:49622), peer Ok(127.0.0.1:1234)
// server: 127.0.0.1:49622, local Ok(127.0.0.1:1234), peer Ok(127.0.0.1:49622)

// 0s | 192.168.1.42
// client: local Ok(192.168.1.42:49716), peer Ok(192.168.1.42:1234)
// server: 192.168.1.42:49716, local Ok(192.168.1.42:1234), peer Ok(192.168.1.42:49716)

// 127.0.0.1 | 192.168.1.42
// Error: Os { code: 61, kind: ConnectionRefused, message: "Connection refused" }
```

The first two work as expected, including setting the correct `local|peer_addr` on each side of the stream. The last two cause panics today due to holes in the stop-gap implementation for loopback.

We need to address this with workarounds and/or include this in the refactor being discussed in #132 .

jeremymill:
This should resolve #135.

Current workaround for loopback from #118 does not allow connecting to the same host when using the public ip. This change extends the workaround to allow for that, unblocking additional use cases pending a larger rework to fully support multiple interfaces (see #132)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -417,6 +417,11 @@ pub fn matches(bind: SocketAddr, dst: SocketAddr) -> bool {
     bind == dst
 }
 
+/// Returns whether loopback is supported from src to dst
+pub(crate) fn is_loopback(src: SocketAddr, dst: SocketAddr) -> bool {
+    dst.ip().is_loopback() || src.ip() == dst.ip()
+}
+
 #[cfg(test)]
 mod test {
     use crate::{Host, Result};
--- a/src/host.rs
+++ b/src/host.rs
@@ -417,8 +417,9 @@ pub fn matches(bind: SocketAddr, dst: SocketAddr) -> bool {
     bind == dst
 }
 
-/// Returns whether loopback is supported from src to dst
-pub(crate) fn is_loopback(src: SocketAddr, dst: SocketAddr) -> bool {
+/// Returns true if loopback is supported between two addresses, or
+/// if the IPs are the same (in which case turmoil treats it like loopback)
+pub(crate) fn is_same(src: SocketAddr, dst: SocketAddr) -> bool {
     dst.ip().is_loopback() || src.ip() == dst.ip()
 }
 
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -17,6 +17,7 @@ use tokio::{
 
 use crate::{
     envelope::{Envelope, Protocol, Segment, Syn},
+    host::is_loopback,
     host::SequencedSegment,
     net::SocketPair,
     world::World,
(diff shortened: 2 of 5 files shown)


src/net/tcp/stream.rs
```diff
@@ -74,7 +74,7 @@ impl TcpStream {
             let rx = host.tcp.new_stream(pair);
 
             let syn = Protocol::Tcp(Segment::Syn(Syn { ack }));
-            if !dst.ip().is_loopback() {
+            if !dst.ip().is_loopback() && dst.ip() != local_addr.ip() {
```
atreyas, line 77: Maybe introduce a function that can compare the host for 2 IPs.

mh32 (approved):

src/host.rs
```diff
@@ -417,6 +417,11 @@ pub fn matches(bind: SocketAddr, dst: SocketAddr) -> bool {
     bind == dst
 }
 
+/// Returns whether loopback is supported from src to dst
```
atreyas, line 420: Technically this is not a loopback. It ends up behaving like loopback because of how turmoil is routing to the local IP address. Mayeb `is_same`?

tests/tcp.rs
```diff
@@ -830,6 +832,38 @@ fn loopback_to_localhost_v4() -> Result {
     run_localhost_test(IpVersion::V4, bind_addr, connect_addr)
 }
 
+#[test]
+fn loopback_wildcard_public_v4() -> Result {
+    let bind_addr = SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), 1234);
+    let connect_addr = SocketAddr::from((Ipv4Addr::new(192, 168, 0, 1), 1234));
+    run_localhost_test(IpVersion::V4, bind_addr, connect_addr)
```
mcches, line 839: Does this verify the stream's addrs are set the same as what tokio::net types return? These need to behave the same way.

mcches (commented):
Thanks for picking up this issue. I left one comment.

tests/tcp.rs
jeremymill: `run_localhost_test` asserts:

```
assert_eq!(socket.local_addr()?.ip(), connect_addr.ip());
assert_eq!(socket.peer_addr()?.ip(), connect_addr.ip());
```

on both server and client.

PR #142, opened 2023-10-20
Fix duplicate FIN in the drop glue

mcches:
If the stream is shutdown manually it sends a FIN on the write half,
however we don't check the state in the drop implementation, which led
to a duplicate FIN being sent.

To fix, `is_shutdown` is checked before sending the FIN.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 [dev-dependencies]
 doc-comment = "0.3.3"
 regex = "1"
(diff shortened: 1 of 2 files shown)


mcches:
Verified 2x of these before, only one after the change.

```txt
2023-10-20T16:36:05.204647Z TRACE node{name="client"}: turmoil: Send src=192.168.0.2:49152 dst=192.168.0.1:1738 protocol=TCP FIN
```

tereshch-aws (approved):

marcbowes (approved):

PR #145, opened 2023-11-03
Fix ephemeral port leak upon tcp stream shutdown

brightcoder:
Do not send a Fin upon the write half drop if the stream is already shutdown but still perform the socket pair clean up.


mcches (approved):
Lgtm!

PR #146, opened 2023-11-03
Fix subtraction overflow bug with min latency

Benjscho:
Previously if you specify a `min_message_latency` greater than the default `max_message_latency`, the test would fail with a tokio error: `overflow when subtracting durations`. This updates the `min_message_latency` setter to update the `max_message_latency` if it is less than the new minimum value.

I also updated the max message latency setter to panic if provided with a value less than the minimum message latency. Given this is much more obviously incorrect to a user of the crate (given the default min latency of 0), I felt a panic was appropriate, and would provide an easier error to debug than the tokio message.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -69,14 +69,34 @@ impl Builder {
             .as_mut()
             .expect("`Latency` missing")
             .min_message_latency = value;
+        // If the min latency is greater than the max latency, update 
+        // the max latency.
+        if self.link
+            .latency
+            .as_mut()
+            .expect("`Latency` missing")
+            .max_message_latency < value {
+                self.link
+                    .latency
+                    .as_mut()
+                    .expect("`Latency` missing")
+                    .max_message_latency = value.clone()
+            }
         self
     }
 
     pub fn max_message_latency(&mut self, value: Duration) -> &mut Self {
+        if self.link
+            .latency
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -69,14 +69,40 @@ impl Builder {
             .as_mut()
             .expect("`Latency` missing")
             .min_message_latency = value;
+        // If the min latency is greater than the max latency, update
```
mcches, line 72: I think it'd be simpler to just assert the min/max are valid on build and not bump up the max here as well.

tests/tcp.rs
```diff
@@ -1041,3 +1041,33 @@ fn socket_to_nonexistent_node() -> Result {
     });
     sim.run()
 }
+
+#[test]
+fn high_host_latency() -> Result {
```
mcches, line 1046: I'm pretty comfortable with this existing test: `override_link_latency`. Did you see some gaps that are covered by this new one?

Benjscho: Ahh I missed that test, no that definitely covers it!

src/builder.rs
Benjscho: Yeah that sounds better, I'll cancel this change and file an updated PR!

PR #147, opened 2023-11-04
Fix subtraction overflow bug with latency

Benjscho:
Previously if you specify a `min_message_latency` greater than the default `max_message_latency`, the test would fail with a tokio error: `overflow when subtracting durations`. This change updates the builder to assert that the min/max latencies are valid on build.


mcches (approved):
Lgtm!

PR #150, opened 2023-11-16
Fix builder to use provided rng

Benjscho:
This commit fixes the turmoil sim builder by removing the `rng`
function. Currently the `build` function ignores the `rng` set on the
builder, and due to the boxing of the value there is no easy way to
repeatedly build a sim with the same `rng`. We would need to add the
ability to clone the boxed rng, which requires some complicated trait
bounds and a different breaking of the interface. This is simpler and
fixes the bug where rng was ignored during building.

This is a breaking change:
- `rng` is removed from the builder, as it was not delivering its
  expected behaviour. Users now need to explicitly specify
  `build_with_rng`.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -5,8 +5,6 @@ use std::time::{Duration, SystemTime};
 
 /// Configure the simulation
 pub struct Builder {
-    rng: Option<Box<dyn RngCore>>,
-
     config: Config,
 
     ip_version: IpVersion,
@@ -23,7 +21,6 @@ impl Default for Builder {
 impl Builder {
     pub fn new() -> Self {
         Self {
-            rng: None,
             config: Config::default(),
             ip_version: IpVersion::default(),
             link: config::Link {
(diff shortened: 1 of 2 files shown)


src/builder.rs
```diff
@@ -93,11 +84,15 @@ impl Builder {
...
 
+    /// Build with default rng.
     pub fn build<'a>(&self) -> Sim<'a> {
         self.build_with_rng(Box::new(rand::rngs::SmallRng::from_entropy()))
     }
 
-    pub fn build_with_rng<'a>(&self, rng: Box<dyn RngCore>) -> Sim<'a> {
+    /// Build a sim with a provided `rng`.
+    ///
+    /// This allows setting the random number generator used to fuzz
```
mcches, line 94:
```suggestion
    /// This allows setting the random number generator used to fuzz.
```

mcches (approved):
Approved with one minor doc nit.

src/builder.rs
```diff
@@ -93,11 +84,15 @@ impl Builder {
...
+    fn build_with_rng<'a>(&self, rng: Box<dyn RngCore>) -> Sim<'a> {
```
cfcosta, line 95: Hey, just so you know, the function was public before and isn't anymore, is that by design?

Benjscho: No, this was a mistake. It should be fixed after #152 is merged

PR #152, opened 2023-11-16
Add docs for the builder and config focused on defaults

Benjscho:
Add documentation to the sim builder and the config, particularly focused on default values.

Also fix a few errors that emerged from running `cargo doc`.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -137,7 +137,7 @@ impl Builder {
         self
     }
 
-    /// The repair rate of connections on the network. This is how
+    /// The repair rate of messages on the network. This is how
     /// frequently a link is repaired after breaking.
     pub fn repair_rate(&mut self, value: f64) -> &mut Self {
         self.link.message_loss_mut().repair_rate = value;
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -3,7 +3,67 @@ use crate::*;
 use rand::{RngCore, SeedableRng};
 use std::time::{Duration, SystemTime};
 
-/// Configure the simulation
+/// A builder that can be used to configure the simulation.
+///
+/// The Builder allows you to set a number of options when creating a turmoil
+/// simulation, see the available methods for documentation of each of these
+/// options.
+///
+/// ## Examples
+///
+/// You can use the builder to initialize a sim with default configuration:
+///
+/// ```
+/// let sim = turmoil::Builder::new().build();
+/// ```
+///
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -54,45 +114,57 @@ impl Builder {
...
     }
 
+    /// The maximum latency that a message will take to transfer over a
+    /// link on the network.
     pub fn max_message_latency(&mut self, value: Duration) -> &mut Self {
         self.link.latency_mut().max_message_latency = value;
         self
     }
 
+    /// The failure rate of connections on the network. This is how
```
mcches, line 131: These are per message so the link is not actually broken. Think of packet loss. It happens to break a tcp connection since we don't have re-send capabilities yet. With UDP, this is pretty useful for testing network flakiness.

src/builder.rs
```diff
@@ -54,45 +114,57 @@ impl Builder {
...
+    /// frequently a link is repaired after being broken.
     pub fn repair_rate(&mut self, value: f64) -> &mut Self {
         self.link.message_loss_mut().repair_rate = value;
         self
     }
 
+    /// Capacity of TCP connections in the sim.
```
mcches, line 145: _a host's tcp buffer_

mcches (commented):
This is great. A couple minor comments.

src/builder.rs
```diff
@@ -3,7 +3,67 @@ use crate::*;
...
-/// Configure the simulation
+/// A builder that can be used to configure the simulation.
+///
+/// The Builder allows you to set a number of options when creating a turmoil
+/// simulation, see the available methods for documentation of each of these
+/// options.
+///
+/// ## Examples
+///
+/// You can use the builder to initialise a sim with default configuration:
```
mcches, line 14:
```suggestion
/// You can use the builder to initialize a sim with default configuration:
```

src/builder.rs
```diff
@@ -54,45 +114,59 @@ impl Builder {
...
+    /// The failure rate of messages on the network. For TCP
+    /// this will break connections as currently there are
+    /// no re-send capabilities. With UDP this is useful for
+    /// testing network flakiness.
     pub fn fail_rate(&mut self, value: f64) -> &mut Self {
         self.link.message_loss_mut().fail_rate = value;
         self
     }
 
+    /// The repair rate of connections on the network. This is how
```
mcches, line 140:
```suggestion
    /// The repair rate of messages on the network. This is how
```

PR #154, opened 2023-11-29
Update Axum example to use axum 0.7 and hyper 1.

TannerRogalsky:
The internals of the various components of this example have changed significantly. While this works, I'd appreciate a careful review to ensure that it is correct in addition to functional.

Thank you!

### DIFF
--- a/examples/axum/Cargo.toml
+++ b/examples/axum/Cargo.toml
@@ -5,11 +5,15 @@ edition = "2021"
 publish = false
 
 [dependencies]
-axum = "0.6"
-hyper = { version = "0.14", features = ["full"] }
-async-stream = "0.3"
+axum = "0.7"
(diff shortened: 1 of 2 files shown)


TannerRogalsky:
The hyper-util dependency has been updated to the most recent release. I believe this is good to merge now baring any review change requests.

getong:
@TannerRogalsky 
How about taking a look at the grpc example, which might be updated too?

TannerRogalsky:
@getong I'm not interested in doing that but you should be able to use my code an example of the things that need to change.

getong:
> @getong I'm not interested in doing that but you should be able to use my code an example of the things that need to change.

I am willing to do this, I tried but failed, I does not know how to fix it.

seanmonstar (approved):

PR #155, opened 2023-11-29
Add documentation on using tracing

Benjscho:
Add some documentation on how to use the tracing level of turmoil to see individual packet flow between clients. This uses the axum-example, since that is already instrumented with a tracing-subscriber.


src/lib.rs
```diff
@@ -60,9 +60,40 @@
 //! # Tracing
 //!
 //! The `tracing` crate is used to emit important events during the lifetime of
-//! a simulation.
+//! a simulation. To enable traces, your tests must be wrapped in a
```
mcches, line 63:
```suggestion
//! a simulation. To enable traces, your tests must install a
```

src/lib.rs
```diff
@@ -60,9 +60,40 @@
...
+//! 2023-11-29T20:23:43.277324Z  INFO client: axum_example: Got response: Response { status: 200, version: HTTP/1.1, headers: {"content-type": "text/plain; charset=utf-8", "content-length": "10", "date": "Wed, 29 Nov 2023 20:23:43 GMT"}, body: b"Hello foo!" }
+//! ...
+//! ```
+//!
+//! Here the server is sending a response, we can see a few steps of the simulation
+//! happen before that response is delivered to the client, received, and then
+//! the client processes the response.
```
mcches, line 96: Consider restructuring this to call attention to the tracing event messages: Sent, Delivered, Recv.

mcches (commented):
Thanks for doing this. A couple minor suggestions and this can merge.

src/lib.rs
Benjscho: Sounds good, just updated. Let me know if anything's inaccurate

src/lib.rs
```diff
@@ -60,9 +60,42 @@
...
+//! 2023-11-29T20:23:43.277324Z  INFO client: axum_example: Got response: Response { status: 200, version: HTTP/1.1, headers: {"content-type": "text/plain; charset=utf-8", "content-length": "10", "date": "Wed, 29 Nov 2023 20:23:43 GMT"}, body: b"Hello foo!" }
+//! ...
+//! ```
+//!
+//! Here the server is sending a response, before it is delivered to, and
+//! received by the client. Note that there are three steps to each packet
+//! trace in turmoil. We see `Send` when a packet is sent from one address
+//! to another. The packet is then `Delivered` to its destination, and when
+//! the destination listener reads the packet it is `Recv`'d.
```
mcches, line 98:
```suggestion
//! the destination reads the packet it is `Recv`'d.
```

mcches (approved):

PR #156, opened 2023-12-07
Pull visibilty policy up

brandonpike:
Before this commit, all visibility was handled at the lowest level possible. This litters the code with pub(crate) in almost every file. Instead, what my commit does is let the highest level possible deal with it.


src/sim.rs
```diff
@@ -34,7 +34,7 @@ pub struct Sim<'a> {
 }
 
 impl<'a> Sim<'a> {
-    pub(crate) fn new(config: Config, world: World) -> Self {
+    pub fn new(config: Config, world: World) -> Self {
```
mcches, line 37: I don't believe this one should change. Construction should happen through the builder.

mcches (commented):
Thanks for doing this. I left one comment that needs to be addressed. Is there any way to diff the rustdoc or something to ensure we didn't break or open up more of the public api surface with this change?

mcches:
Closing due to inactivity.

PR #157, opened 2023-12-20
Add a check for port exhaustion

mcches:
I never thought we'd run through the whole port range, but it seems
some complex setups can hit this.

Add a panic if we run out of ports.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -75,9 +75,9 @@ impl Host {
         for _ in PORT_RANGE {
             let ret = self.next_ephemeral_port;
 
-            if self.next_ephemeral_port == *PORT_RANGE.start() {
+            if self.next_ephemeral_port == *PORT_RANGE.end() {
                 // re-load
-                self.next_ephemeral_port = *PORT_RANGE.end();
+                self.next_ephemeral_port = *PORT_RANGE.start();
             } else {
                 // advance
                 self.next_ephemeral_port += 1;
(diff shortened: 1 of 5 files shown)


src/host.rs
```diff
@@ -9,10 +9,13 @@ use std::collections::VecDeque;
 use std::fmt::Display;
 use std::io;
 use std::net::{IpAddr, SocketAddr};
+use std::ops::RangeInclusive;
 use std::sync::Arc;
 use tokio::sync::{mpsc, Notify};
 use tokio::time::{Duration, Instant};
 
+const PORT_RANGE: RangeInclusive<u16> = 49152..=65535;
```
marcbowes, line 17: Perhaps make this configurable

tests/tcp.rs
```diff
@@ -1134,3 +1134,31 @@ fn socket_to_nonexistent_node() -> Result {
...
     sim.run()
 }
+
+#[should_panic]
+#[test]
+fn exhaust_ephemeral_ports() {
+    let mut sim = Builder::new()
+        .tick_duration(Duration::from_millis(1))
+        .simulation_duration(Duration::from_secs(60))
+        .build();
```
marcbowes, line 1144: If the range was configurable, this test could be "faster" if you gave it a smaller range.

marcbowes (approved):

mh32 (approved):

PR #161, opened 2023-12-22
Lucio/spawn blocking bug

LucioFranco:


mcches:
This is reference code. We've been able to reproduce and identify the issue when `spawn_blocking` is used within turmoil. The problem here is that time does not advance on a paused runtime when a blocking task is running, which locks up the sim's step loop.

PR #162, opened 2024-01-05
Add a tracing warning when a thread blocks steps

### LINKED ISSUE #160: Add warning for blocking tasks that block the sim
Related to #139 we should add a warning that prints when a blocking task is still active in the runtime causing the next tick to not happen. This can be done by 1) adding a blocking task count metric to `tokio-metrics` and then to spin a bg thread that checks this metric and some sort of tick count. It will then start printing if the tick can not progress.

Benjscho:
Add a warning to the sim when a given host or client blocks progress in a simulation run. This works by spawning a background thread for each run that periodically checks the steps taken by the simulation. If the number of steps is the same between checks then the thread adds the tracing info.

Few questions here:
- Is 10s the right amount of time? Should this be configurable? 
- Is tracing warning the right way of displaying this info? Any print warnings are swallowed when running `cargo test` without `--nocapture`

Fixes #160

### DIFF
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -5,7 +5,10 @@ use std::cell::RefCell;
 use std::future::Future;
 use std::net::IpAddr;
 use std::ops::DerefMut;
-use std::sync::Arc;
+use std::sync::mpsc::TryRecvError;
+use std::sync::{Arc, mpsc};
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering::Relaxed;
 use std::time::UNIX_EPOCH;
 use tokio::time::Duration;
 use tracing::Level;
@@ -30,7 +33,7 @@ pub struct Sim<'a> {
     /// Simulation elapsed time
     elapsed: Duration,
 
-    steps: usize,
+    steps: Arc<AtomicUsize>,
 }
 
 impl<'a> Sim<'a> {
@@ -46,7 +49,7 @@ impl<'a> Sim<'a> {
             rts: IndexMap::new(),
             since_epoch,
             elapsed: Duration::ZERO,
-            steps: 1, // bumped after each step
+            steps: Arc::new(1.into()), // bumped after each step
         }
     }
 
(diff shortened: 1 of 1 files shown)


Benjscho:
@LucioFranco would you be able to review?

src/sim.rs
```diff
@@ -380,12 +400,12 @@ impl<'a> Sim<'a> {
...
 
         self.elapsed += tick;
-        self.steps += 1;
+        self.steps.fetch_add(1, Relaxed);
 
         if self.elapsed > self.config.duration && !is_finished {
             return Err(format!(
                 "Ran for duration: {:?} steps: {} without completing",
-                self.config.duration, self.steps,
+                self.config.duration, self.steps.load(Relaxed),
```
LucioFranco, line 408: You can avoid this load because fetch_add above returns the previous value so you can just +1 that and get the current value.

src/sim.rs
```diff
@@ -309,10 +312,27 @@ impl<'a> Sim<'a> {
...
+        let (tx, rx) = mpsc::channel();
+        std::thread::spawn(move || {
+            loop {
+                let prev = steps.load(std::sync::atomic::Ordering::Relaxed);
+                // Exit if main thread has.
+                match rx.try_recv() {
+                    Ok(_) | Err(TryRecvError::Disconnected) => break,
+                    _ => {}
+                }
+                std::thread::sleep(Duration::from_secs(10));
```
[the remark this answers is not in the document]
Benjscho: That's a good point, we could log it only once per step with the step it's stuck on and then skip the log after that. I'd think something's likely broken if a single step is taking 10s of real time work

src/sim.rs
Benjscho: Nice, will adjust that!

src/sim.rs
```diff
@@ -309,10 +312,33 @@ impl<'a> Sim<'a> {
...
     /// returning early if any host software errors.
     pub fn run(&mut self) -> Result {
+        let steps = self.steps.clone();
+        let (tx, rx) = mpsc::channel();
+        std::thread::spawn(move || {
+            let mut blocked = false;
+            loop {
+                let prev = steps.load(std::sync::atomic::Ordering::Relaxed);
+                // Exit if main thread has.
+                match rx.try_recv() {
```
mcches, line 322: How does this behave when you call run() N times in a row? It looks like you could spawn a ton of threads that don't clean up for 10s.

src/sim.rs
```diff
@@ -309,10 +312,33 @@ impl<'a> Sim<'a> {
...
+                } else {
+                    blocked = false;
+                }
+            }
+        });
         loop {
             let is_finished = self.step()?;
 
             if is_finished {
+                let _ = tx.send(());
```
mcches, line 341: Doesn't the drop handle this for you?

mcches (commented):
Did you a consider a strategy where this is fallible? If X duration of real time elapses and the sim doesn't progress, fail the test. This saves operators from cancelling run away builds.

mcches:
> Did you a consider a strategy where this is fallible? If X duration of real time elapses and the sim doesn't progress, fail the test. This saves operators from cancelling run away builds.

I'll step back on this, and instead we can expect CI systems running these to set the timeout to be less oppinionated. For example, nextest is pretty great for configuring this.

src/sim.rs
Benjscho: That's a good point, I'll adjust it so it uses `recv_timeout` - that way if sim exits early the background thread will be closed too and it'll clean up straight away

src/sim.rs
Benjscho: Good point

Benjscho:
@mcches or @LucioFranco would you mind giving this another look when you have a min? I've addressed all prev comments

Benjscho:
Discussed offline as planning to not implement. Blocking warning would be nice, but most test runners (e.g., nextest) provide warnings for long running tests. Spawning a background thread per test for parallel scenarios seemed incorrect

PR #163, opened 2024-01-07
Update actions/checkout action and replace unmaintained actions-rs actions

taiki-e:
- Update actions/checkout action
- Use cargo directly instead of unmaintained actions-rs/cargo action
- Use [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) instead of unmaintained actions-rs/toolchain action

See also https://github.com/tokio-rs/tokio/pull/5316 for the context of unmaintained actions and alternatives.


mcches (approved):
Thank you!

PR #164, opened 2024-01-08
Add `turmoil::sim_elapsed` for retrieving total simulation virtual time

zakvdm:
Previously, we only exposed `turmoil::elapsed` which returned the elapsed virtual time for the currently executing host. If you step your simulation for some duration before registering a host then the host's elapsed time will be less than the overall sim elapsed time.

We didn't want to make a breaking API change for this, so opted to add a new function (`turmoil::sim_elapsed`). This function returns an `Option` so that the caller can gracefully handle failure (`turmoil::elapsed` will crash if called outside of an executing host, for example). This makes it easier to use `sim_elapsed` for things like logging elapsed time via the std tracing subscriber (there's an example of how to do this in the `grpc` example).

One sharp edge I ran into is that we sometimes do tracing in this crate itself (e.g. in `Tcp::bind`) which happens while we're holding a mutable borrow of the thread-local `RefCell<World>`. Since the tracing event causes us to call `tracing::sim_elapsed`, we have an attempted double mutable borrow. In these cases I just made `sim_elapsed` return `None`. It's a bit of a wart but I think it's okay. We also return `None` when `sim_elapsed` is called when there's no executing host, so I think the API returning `Option` feels right. We can internally fix the double borrow issue in future without a breaking change.

### DIFF
--- a/examples/grpc/src/main.rs
+++ b/examples/grpc/src/main.rs
@@ -19,11 +19,7 @@ use proto::{HelloReply, HelloRequest};
 use crate::proto::greeter_client::GreeterClient;
 
 fn main() {
-    if std::env::var("RUST_LOG").is_err() {
-        std::env::set_var("RUST_LOG", "info");
-    }
-
-    tracing_subscriber::fmt::init();
+    configure_tracing();
 
     let addr = (IpAddr::from(Ipv4Addr::UNSPECIFIED), 9999);
 
@@ -60,7 +56,7 @@ fn main() {
             let request = Request::new(HelloRequest { name: "foo".into() });
             let res = greeter_client.say_hello(request).await?;
 
-            tracing::info!("Got response: {:?}", res);
+            tracing::info!(?res, "Got response");
 
             Ok(())
         }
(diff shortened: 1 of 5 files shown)


examples/grpc/src/main.rs
```diff
@@ -70,6 +66,32 @@ fn main() {
...
+    .expect("Configure tracing");
+}
+
+#[derive(Clone)]
+struct SimElapsedTime;
+impl tracing_subscriber::fmt::time::FormatTime for SimElapsedTime {
+    fn format_time(&self, w: &mut tracing_subscriber::fmt::format::Writer<'_>) -> std::fmt::Result {
+        // The debug implementation of Duration gives reasonably formatted
+        // durations (e.g. `610ms`, `2.62s`).
+        write!(w, "{:?}", turmoil::sim_elapsed().unwrap_or_default())
```
mcches, line 91: I think it is valuable to see real and sim time in the trace. Up to you on whether you think that is useful in the example.

src/lib.rs
```diff
@@ -170,6 +172,23 @@ pub(crate) fn for_pairs(a: &Vec<IpAddr>, b: &Vec<IpAddr>, mut f: impl FnMut(IpAd
     }
 }
 
+/// Returns how long the currently executing host has been executing for in
+/// virtual time.
+///
+/// Must be called from within a Turmoil simulation.
+pub fn elapsed() -> Duration {
```
mcches, line 179: While the docs are the same, this is definitely more readable in the source code.

mcches (approved):
Nice change!

PR #165, opened 2024-01-08
Bump tokio to version "1"

mcches:
Opting to only specify major version to limit dependency maintenance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ rand = { version = "0.8.5", features = ["small_rng"] }
 rand_distr = "0.4.3"
 regex = { version = "1", optional = true }
 scoped-tls = "1.0.1"
-tokio = { version = "1.25.0", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
 tokio-stream = "0.1"
 tokio-test = "0.4.2"
 tokio-util = "0.7.4"


Cargo.toml
```diff
@@ -27,7 +27,7 @@ rand = { version = "0.8.5", features = ["small_rng"] }
 rand_distr = "0.4.3"
 regex = { version = "1", optional = true }
 scoped-tls = "1.0.1"
-tokio = { version = "1.25.0", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
```
seanmonstar, line 30: Can it work with "earlier" Tokio versions? If not, keeping this set to the minimum version that it can work with is usually best. Can be checked with `--minimal-versions`.

mcches: I don't have nightly. Is there a stable way to check this?

I actually looked at `hyper` before changing this. Is hyper >= 1.0.0 only?

seanmonstar: hyper only depends on some methods that existed at 1.0. I think Tokio needed to add something for turmoil, no? In that case, you'd need that version as the minimum one here.

seanmonstar: Oh I didn't mean it knows about turmoil, but that something was missing before turmoil could exist. Maybe I'm wrong.

You could also add a minimal versions job to the CI. Or if you're certain there's nothing, we can just ship. Up to you.

mcches: I see. So we use [`unhandled_panic`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.unhandled_panic), which is newer than 1.0.0. I think I'll cancel this for now. Thanks for the review!

mcches:
Cancelling.

PR #170, opened 2024-02-16
Make sim_elapsed function safe to call outside of simulation

tereshch-aws:
- **Make sim_elapsed function safe to call outside of simulation**


zakvdm:
Thanks for this, looks good to me.

zakvdm (approved):

PR #171, opened 2024-03-26
Enable tokio io driver

HolyShitMan:
**Pull request includes:**

* activation of tokio io driver to  enable _net_, _process_ & _signals_ within runtime (for our use-case we need _net_)
* some additional formating stuff

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,10 @@
 use crate::*;
 
 use rand::{RngCore, SeedableRng};
-use std::{ops::RangeInclusive, time::SystemTime};
+use std::{
+    ops::RangeInclusive,
+    time::{Duration, SystemTime},
+};
 
 /// A builder that can be used to configure the simulation.
 ///
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -164,6 +164,7 @@ impl Builder {
         self
     }
 
+    /// Enables the tokio I/O driver.
     pub fn enable_tokio_io(&mut self) -> &mut Self {
         self.config.enable_tokio_io = true;
         self
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,10 +1,8 @@
-use crate::*;
+use std::{ops::RangeInclusive, time::SystemTime};
 
 use rand::{RngCore, SeedableRng};
-use std::{
-    ops::RangeInclusive,
-    time::{Duration, SystemTime},
-};
+
+use crate::*;
 
 /// A builder that can be used to configure the simulation.
 ///
@@ -163,6 +161,16 @@ impl Builder {
     /// Capacity of host's UDP buffer in the sim.
     pub fn udp_capacity(&mut self, value: usize) -> &mut Self {
         self.config.udp_capacity = value;
+        self
+    }
+
+    pub fn enable_tokio_io(&mut self) -> &mut Self {
+        self.config.enable_tokio_io = true;
+        self
+    }
+
+    pub fn disable_tokio_io(&mut self) -> &mut Self {
+        self.config.enable_tokio_io = false;
         self
     }
 
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -169,11 +169,6 @@ impl Builder {
         self
     }
 
-    pub fn disable_tokio_io(&mut self) -> &mut Self {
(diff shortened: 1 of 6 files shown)


mcches:
This has come up a couple of times in the past. We intentionally don't enable io to avoid confusion between tokio::net and turmoil::net types. Can I ask why you aren't using turmoil::net types in your tests and for a little more information about your use case?

HolyShitMan:
We certainly understand this intention. We have currently two use-cases, that use the tokio io types:
1. Manual GUI testing framework that uses turmoil internally and exposes one socket externally to attach GUI.
2. Tests that write files onto the HDD (missed this in my initial comment onto the PR)

Maybe a compromise could be to hide the IO driver activation behind a feature flag?!

mcches:
> 1. Manual GUI testing framework that uses turmoil internally and exposes one socket externally to attach GUI.

Interesting! I'd love to hear more about this or see a demo.

> 2. Tests that write files onto the HDD (missed this in my initial comment onto the PR)

This should be fine if you use tokio fs as that doesn't use the io reactor.

> Maybe a compromise could be to hide the IO driver activation behind a feature flag?!

I'd be more comfortable doing it like this.

mcches:
You could also smuggle messages out via a channel and then run your socket independently of turmoil.

HolyShitMan:
> You could also smuggle messages out via a channel and then run your socket independently of turmoil.

This of course could also be possible. 
Any strong opinion from your side, whether to include this PR?

mcches:
> > You could also smuggle messages out via a channel and then run your socket independently of turmoil.
> 
> This of course could also be possible. Any strong opinion from your side, whether to include this PR?

If it unblocks you and it isn't too much of a hassle I'd prefer that approach. One reason for this is that at some point down the line we'd like to incorporate this into tokio itself, at which point you'd set a flag on the runtime and you get one or the other. I'm not sure when that will be, but if we can avoid it now that would save you some upgrade headache later.

HolyShitMan:
I talked to all my colleges, that are involved in this issue of needing IO within unit tests, and the consensus was, that by not using the tokio IO driver the short and medium term effort is to high and our code base readability decreases massively.
Still, we fully understand your point, and if you're not inclined to merge this PR, we propose two alternatives:

1. I withdraw this PR and create a new one, introducing an additional constructor that accepts a user-created tokio runtime as a parameter. Maybe this aligns better with your long-term goals.
2. If this isn't feasible, we'll continue using our local turmoil fork

mcches:
Let's do a Builder option and proceed with enabling it. This has the advantage of being very explicit and allows for tests to have it on and off in the same crate.

HolyShitMan:
Not completely sure about everything, e.g., building no_software runtime with ` Config::default();`, yet count this as a first version to discuss about.

src/builder.rs
```diff
@@ -166,6 +164,16 @@ impl Builder {
         self
     }
 
+    pub fn enable_tokio_io(&mut self) -> &mut Self {
+        self.config.enable_tokio_io = true;
+        self
+    }
+
+    pub fn disable_tokio_io(&mut self) -> &mut Self {
```
mcches, line 172: nit: I don't think we'll need to switch a builder back and forth so one method seems fine. If you have a need for it consider using a single method that takes a bool.

src/rt.rs
```diff
@@ -48,14 +51,17 @@ pub(crate) struct Rt<'a> {
     /// Optional handle to a host's software. When software finishes, the handle is
     /// consumed to check for error, which is propagated up to fail the simulation.
     handle: Option<JoinHandle<Result>>,
+
+    /// Configuration of simulation
+    sim_cfg: Config,
```
mcches, line 56: If you switch to just the enable_io boolean here you don't need to worry about a default cfg for the no software case.

mcches (commented):
Looks pretty good! A couple minor comments.

src/rt.rs
mcches: I don't see this in the recent changes.

mcches (approved):
Lgtm!

mcches:
Merged. I will release this by eow.

HolyShitMan:
thx for your work

PR #173, opened 2024-04-23
Add ability to run hosts in random order

progwriter:
This was quite useful in our tests, where the outcome of a multi-node simulation is sensitive to processing order.

Plumbed through as a boolean option through Config/Builder, w/ the default being `false` (current behavior)


mcches (approved):
lgtm!

PR #174, opened 2024-04-23
Add API parity stubs for UDP sockets/TCP streams

progwriter:
These are no-ops in turmoil (AFAICT, based on patterns in our tests), but having these available minimizes the amount of caller code that needs to change

- https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.set_nodelay
- https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.join_multicast_v6
- https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.set_multicast_loop_v6


progwriter:
Done

mcches (approved):
lgtm!

PR #175, opened 2024-07-03
Make `Config.duration` as `Option<Duration>`

quettabit:
i think this could be handy for users who'd like to have a custom loop containing `sim.step()` and other fault calls. for such custom loops, i think running for a fixed number of steps would be the only way to go. currently, for such cases, they should also be figuring out and configuring the simulation duration appropriately to go well with their simulation, so they don't face the `"Ran for duration: {:?} steps: {} without completing"` error. this change will enable them to just pass `None` for it.

pls let me know if this kind of change is acceptable.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -25,7 +25,7 @@ use crate::*;
 /// use std::time::{Duration, SystemTime};
 ///
 /// let sim = turmoil::Builder::new()
-///     .simulation_duration(Duration::from_secs(60))
+///     .simulation_duration(Some(Duration::from_secs(60)))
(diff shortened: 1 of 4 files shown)


mcches:
Thanks for cutting a PR. Given this is a breaking API change I think we should consider alternatives. Passing [Duration::MAX](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX) is about as much lift as passing `None`, and given that using `step()` manually is more of a super user feature this seems like an OK workaround. wdyt?

quettabit:
cool. yeah makes sense. yeah, that workaround is fine.

PR #176, opened 2024-07-12
Add try_write to TcpStream

brandonpike:
Add `TcpStream::try_write()` and `WriteHalf::try_write()`. 

These allow for a result-based synchronous write attempt.

### DIFF
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -106,41 +106,6 @@ impl TcpStream {
     /// If data is successfully written, `Ok(n)` is returned, where `n` is the
     /// number of bytes written. If the stream is not ready to write data,
     /// `Err(io::ErrorKind::WouldBlock)` is returned.
-    ///
-    /// # Examples
-    ///
-    /// ```no_run
-    /// use turmoil::net::TcpStream;
-    /// use std::error::Error;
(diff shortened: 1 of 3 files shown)


src/net/tcp/stream.rs
```diff
@@ -93,6 +93,21 @@ impl TcpStream {
...
+    /// Try to write a buffer to the stream, returning how many bytes were
+    /// written.
+    ///
+    /// The function will attempt to write the entire contents of `buf`, but
+    /// only part of the buffer may be written.
+    ///
+    /// # Return
+    ///
+    /// If data is successfully written, `Ok(n)` is returned, where `n` is the
+    /// number of bytes written. If the stream has been shut down
```
mcches, line 105: nit: I'd just copy the doc verbatim. In future changes this will do what the tokio flavor does and then we don't need to update the doc.

mcches (commented):
Lgtm, please add a test that hits this method.

src/net/tcp/stream.rs
brandonpike: Missed this nit. You're probably right, although I see some (small) benefit in the doc representing the current state

mcches (approved):

Issue #167, opened 2024-01-26
Support one-way partitions


mooso:
Would be great if turmoil can publicly support the ability to introduce one-way partitions between hosts: host A can send messages to host B, but host B messages don't get delivered to host A.

PR #187, opened 2024-09-30
Allow one way partitionings

andersmusikkahs:
This is an attempt to solve #167 .

### DIFF
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -229,7 +229,7 @@ impl<'a> Sim<'a> {
     /// from 'from' to 'to', while not affecting the ability for them to be delivered
     /// in the other direction.
     ///
-    /// Partitioning first a->b, then b->a, will stop all messages, similarly to
+    /// Partitioning first from->to, then to->from, will stop all messages, similarly to
     /// a single call to [`Self::partition`].
     ///
     /// Combining this feature with [`Self::hold`] is presently not supported.
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -203,6 +203,15 @@ impl<'a> Sim<'a> {
         world.repair_many(a, b);
     }
 
+    /// Repair the connection between two hosts, or sets of hosts, removing
+    /// the effect of a previous [`Self::partition_oneway`].
+    ///
+    /// Combining this feature with [`Self::hold`] is presently not supported.
+    pub fn repair_oneway(&self, from: impl ToIpAddrs, to: impl ToIpAddrs) {
+        let mut world = self.world.borrow_mut();
+        world.repair_oneway_many(from, to);
+    }
+
     /// The opposite of [`hold`](crate::hold). All held messages are immediately delivered.
     pub fn release(&self, a: impl ToIpAddrs, b: impl ToIpAddrs) {
         let mut world = self.world.borrow_mut();
@@ -214,6 +223,19 @@ impl<'a> Sim<'a> {
     pub fn partition(&self, a: impl ToIpAddrs, b: impl ToIpAddrs) {
         let mut world = self.world.borrow_mut();
         world.partition_many(a, b);
+    }
+
+    /// Partition two hosts, or sets of hosts, such that messages can not be sent
+    /// from 'from' to 'to', while not affecting the ability for them to be delivered
+    /// in the other direction.
+    ///
+    /// Partitioning first a->b, then b->a, will stop all messages, similarly to
+    /// a single call to [`Self::partition`].
+    ///
+    /// Combining this feature with [`Self::hold`] is presently not supported.
+    pub fn partition_oneway(&self, from: impl ToIpAddrs, to: impl ToIpAddrs) {
+        let mut world = self.world.borrow_mut();
+        world.partition_oneway_many(from, to);
     }
 
     /// Resolve host names for an [`IpAddr`] pair.
@@ -407,10 +429,12 @@ impl<'a> Sim<'a> {
 
 #[cfg(test)]
 mod test {
+    use rand::Rng;
     use std::{
         net::{IpAddr, Ipv4Addr},
         rc::Rc,
         sync::{
+            Mutex,
             Arc,
             atomic::{AtomicU64, Ordering},
         },
@@ -424,10 +448,11 @@ mod test {
         time::Instant,
     };
 
+    use crate::net::UdpSocket;
     use crate::{
         Builder, elapsed,
         hold,
-        net::{TcpListener, TcpStream}, Result, sim_elapsed, World,
+        net::{TcpListener, TcpStream}, Result, Sim, sim_elapsed, World,
     };
 
     #[test]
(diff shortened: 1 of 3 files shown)


src/sim.rs
```diff
@@ -660,6 +683,52 @@ mod test {
...
+                .await
+                .expect("sending packet should work");
+            tokio::time::timeout(Duration::from_secs(1), udp_socket.recv_from(&mut [0]))
+                .await
+                .expect_err("sender should not hear receiver");
+
+            Ok(())
+        });
+
+        sim.partition_oneway("receiver", "sender");
```
mcches, line 726: Could you also add the following tests please:

- verify repair works after partition
- check with two different a and b pairings to ensure sort ordering works as intended

src/sim.rs
```diff
@@ -660,6 +683,52 @@ mod test {
...
+            let mut buf = [0];
+            tokio::time::timeout(Duration::from_secs(1), udp_socket.recv_from(&mut buf))
+                .await
+                .expect("receiver should receive a message before the timeout")
+                .expect("the message should be received successfully");
+            Ok(())
+        });
+
+        sim.client("sender", async move {
+            // Peers are partitioned. TCP setup should fail.
```
mcches, line 713: TCP?

src/top.rs
```diff
@@ -440,7 +482,55 @@ impl Link {
     }
 
     fn explicit_partition(&mut self) {
-        self.state = State::ExplicitPartition;
+        self.state = State::ExplicitPartitionBothWays;
+    }
+
+    fn explicit_partition_a_to_b(&mut self) {
```
mcches, line 488: We'll also need test coverage over these state transitions.

src/top.rs
```diff
@@ -332,14 +368,20 @@ impl Link {
                 let delay = self.delay(global_config.latency(), rand);
                 DeliveryStatus::DeliverAfter(self.now + delay)
             }
+            State::ExplicitPartitionAToB if src.ip() > dst.ip() => {
```
mcches, line 371: This is a bit tricky to reason about as written. I assume this delivers in the non-partitioned link direction. Consider adding a comment on these branches.

mcches (changes requested):
Thanks for picking up this issue!

src/sim.rs
andersmusikkahs: Ooops😬! A bit of copy-paste may have happened here.

src/top.rs
andersmusikkahs: I agree. I'll see if I can make it clearer.

src/top.rs
andersmusikkahs: I added an exhaustive test that all 6 operations that affect partitioning do work as expected in all 6! (=720) combinations.

This gives some peace of mind, but it takes a while to run in debug-mode (about 8 seconds) . One possibility would be to just disable this test for debug-builds. Another possibility would be to not run all 720 combinations.

src/sim.rs
andersmusikkahs: I added tests for repair.

I made the test so that it doesn't always create client "a" first, to try to establish that the sort-orderings don't depend on which client is created first, but I'm not sure if this is what you're asking for!

src/sim.rs
andersmusikkahs: Reference to TCP removed!

src/top.rs
andersmusikkahs: I've tried to make it clearer.

src/top.rs
```diff
@@ -155,7 +155,15 @@ enum State {
     Healthy,
 
     /// The link was explicitly partitioned.
-    ExplicitPartition,
```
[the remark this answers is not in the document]
mcches: I haven't seen a need for this yet, but one idea to achieve this is to model a link with two direction states and allow setting either one. This also might make the A,B -> src,dst stuff cleaner. wdyt?

src/top.rs
mcches: I think we could limit the number of cases down to still be pretty fast, but also give us confidence in the implementation. You could also use a lower level test on this module that doesn't need all the packets to move around or all the host scheduling.

mcches (commented):
This is coming together nicely. I left a few comments, please let me know what you think.

src/top.rs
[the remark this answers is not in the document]
andersmusikkahs: I think it would be a lot cleaner if the 'links'-map in the 'Topology'-struct didn't deduplicate links. I.e, if the network topology between any two nodes 'a' and 'b' was described by two links, 'a'->'b' and 'b'->'a', instead of having one multidirectional link.

However, that feels like a rather big change, and it might also have other downsides.

We could just split 'state' in struct 'Link' into 'state_a_to_b' and 'state_b_to_a'. I worry that the tricky A,B->src,dst stuff would basically remain the same though?

src/top.rs
andersmusikkahs: Done!

src/top.rs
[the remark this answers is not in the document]
mcches: I agree we should hold off on a larger refactor. We've been discussing a larger restructuring of the network to support different ip versions and a more realistic lo for some time now. We can tackle that then. 

This idea doesn't avoid complexity around the directions, but it does make it really clear what is happening in each direction. I think its worth doing that as part of this change.

```rust
struct Link {
    a_b: State,
    b_a: State,
}
```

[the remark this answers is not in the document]
andersmusikkahs: Right! I made an attempt, and I have to say it was a big improvement. Thanks for the idea!

src/sim.rs
```diff
@@ -661,6 +686,207 @@ mod test {
...
+                        Ok(Ok(_))
+                    ));
+
+                    Ok(())
+                }
+            })
+        };
+
+        let construct_a_first = sim.world.borrow_mut().rng.gen_bool(0.5);
+        if construct_a_first {
```
mcches, line 771: What is the significance of the construction order?

src/sim.rs
```diff
@@ -661,6 +686,207 @@ mod test {
...
+            sim.partition_oneway("a", "b");
+            Expectation {
+                expect_a_receive: true,
+                expect_b_receive: false,
+            }
+        })
+    }
+
+    #[test]
+    fn partition_peers_oneway_many_cases() -> Result {
```
mcches, line 803: How's the duration on this now?

mcches (commented):
Mainly just one minor comment about arg names. Looks like you also have a couple clippy violations: https://github.com/tokio-rs/turmoil/actions/runs/11164962511/job/31053944943?pr=187

src/top.rs
```diff
@@ -251,10 +259,38 @@ impl Topology {
         self.links[&Pair::new(a, b)].explicit_partition();
     }
 
+    pub(crate) fn partition_oneway(&mut self, a: IpAddr, b: IpAddr) {
```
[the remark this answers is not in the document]
andersmusikkahs: Yeah, that's obviously much better!

src/sim.rs
andersmusikkahs: It's quite fast.It's about 1/100 of the number of cases compared to before, which really improved performance.

src/top.rs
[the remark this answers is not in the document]
andersmusikkahs: I have implemented your suggestion (I hope :-) ).

mcches:
Looks like this needs a rebase and was hoping we could get `from` and `to` on the pub facing Apis for this.

andersmusikkahs:
> Looks like this needs a rebase and was hoping we could get `from` and `to` on the pub facing Apis for this.

I rebased, squashed into a single commiit, and renamed the parameter 'a' and 'b' to 'from' and 'to'.

src/sim.rs
```diff
@@ -216,6 +225,19 @@ impl<'a> Sim<'a> {
         world.partition_many(a, b);
     }
 
+    /// Partition two hosts, or sets of hosts, such that messages can not be sent
+    /// from 'from' to 'to', while not affecting the ability for them to be delivered
+    /// in the other direction.
+    ///
+    /// Partitioning first a->b, then b->a, will stop all messages, similarly to
```
mcches, line 232:
```suggestion
    /// Partitioning first from->to, then to->from, will stop all messages, similarly to
```

mcches (approved):
I pulled your changes and everything looks good. Thanks for the contribution!

PR #188, opened 2024-10-04
Add established_tcp_stream_count{,_on}

marcbowes:
This commit adds a utility method to count the number of established tcp streams on a given host (or the current host). This can be used to verify connections have been cleaned up, which is sometimes tricky to reason about due to a combination of how tcp cleans up and how Drop works.


mcches (approved):
Thanks for the contribution!

PR #189, opened 2024-10-04
Upgrade gRPC example deps

mcches:


examples/grpc/src/main.rs
```diff
@@ -109,88 +107,81 @@ impl Greeter for MyGreeter {
     }
 }
 
-mod connector {
-    use hyper::Uri;
-    use hyper_util::client::legacy::connect::Connected;
-    use hyper_util::rt::TokioIo;
+mod incoming {
```
mcches, line 110: We could avoid this by implementing in turmoil, but we'd have to take a tonic dependency and managing version upgrades will be difficult.

examples/grpc/src/main.rs
```diff
@@ -109,88 +107,81 @@ impl Greeter for MyGreeter {
...
+
+    use hyper::Uri;
+    use hyper_util::rt::TokioIo;
+
+    use tower::Service;
+    use turmoil::net::TcpStream;
+
+    type Fut = Pin<Box<dyn Future<Output = Result<TokioIo<TcpStream>, std::io::Error>> + Send>>;
+
+    pub fn connector(
```
mcches, line 177: This is pretty clean now.

examples/grpc/src/main.rs
LucioFranco: I think this is fine for something that users can copy into their own code pretty easily and so we dont need to pin turmoil to tonic.

PR #194, opened 2024-11-15
Use compile_fds to compile file descriptor set

tottoto:
`compile_fds` can be used to compile the file descriptor set.


mcches (approved):
Thanks for simplifying! Sorry for the delay.

PR #196, opened 2024-12-26
Use EnvFilter config to set default tracing level

tottoto:
`EnvFilter`'s config can be used to set the default tracing level instead of editing the environment variable.


mcches (approved):
Thanks for improving this!

Issue #184, opened 2024-08-27
Add try_read and friends to TcpStream


mcches:
`try_write` was recently added, but we still need `try_read` and `readable`.

Colin1860:
Is this something that could be considered a good first issue?

mcches:
I had to shelve this and never unassigned, appologies.

PR #206, opened 2024-12-30
Add peek to TcpStream.

nickgian:
Our codebase is using the `peek` functions that Turmoil does not model. It seemed easy to add so here's an attempt to add them for TCP.

- [poll_peek](https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.poll_peek)
- [peek](https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.peek)

### DIFF
--- a/src/net/tcp/split_owned.rs
+++ b/src/net/tcp/split_owned.rs
@@ -36,6 +36,19 @@ impl OwnedReadHalf {
     pub fn reunite(self, other: OwnedWriteHalf) -> Result<TcpStream, ReuniteError> {
         reunite(self, other)
     }
+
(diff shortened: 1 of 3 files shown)


mcches:
I'll be out for a couple of days so my review will be delayed until later in the week. Adding @LucioFranco if you have availability.

I haven't looked yet, but we'll want to ensure #184 fits into whatever is added here.

marcbowes:
I'm taking a look at this now :)

marcbowes (approved):
Thanks for the contribution and comprehensive tests!

PR #211, opened 2025-01-29
implement `UdpSocket::poll_recv_from`

### LINKED ISSUE #209: support `UdpSocket::poll_recv_from`
Tokio's UDP socket has a [poll_recv_from](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.poll_recv_from) method, while turmoil's does not.

m-mueller678:
This fixes #209.

### DIFF
--- a/tests/udp.rs
+++ b/tests/udp.rs
@@ -1,6 +1,5 @@
-use std::future::Future;
-use std::pin::Pin;
-use std::task::{Context, Poll};
+use std::future::poll_fn;
+use std::task::Poll;
 use std::{
     io::{self, ErrorKind},
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
@@ -168,24 +167,15 @@ fn poll_recv() -> Result {
     let mut sim = Builder::new().build();
 
     sim.client("server", async {
-        struct TestFuture<F: FnMut(&mut Context)>(F);
-        impl<F: FnMut(&mut Context) + Unpin> Future for TestFuture<F> {
-            type Output = ();
-
-            fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-                Pin::into_inner(self).0(cx);
-                Poll::Ready(())
-            }
-        }
-
         let expected_origin = lookup("client");
         let sock = bind().await?;
         let buffer = &mut [0u8; 64];
         let mut read_buf = ReadBuf::new(buffer);
 
-        TestFuture(|cx| {
+        poll_fn(|cx| {
             let received = sock.poll_recv_from(cx, &mut read_buf);
             assert!(matches!(received, Poll::Pending));
+            Poll::Ready(())
         })
         .await;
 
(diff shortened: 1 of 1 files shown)


mcches (commented):
Could you add a test that covers pending and ready poll states?

src/net/udp.rs
```diff
@@ -60,6 +64,23 @@ impl Rx {
...
+        &mut self,
+        cx: &mut Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<Result<SocketAddr>> {
+        let (datagram, origin) = if let Some(datagram) = self.buffer.take() {
+            datagram
+        } else {
+            ready!(self.recv.poll_recv(cx)).expect("sender should never be dropped")
+        };
+        buf.clear();
```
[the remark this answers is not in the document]
m-mueller678: yes, `poll_recv` should append to the buffer, not overwrite it. The doc comment should also be updated to reflect that. Tough I don't know when I'll get back to it.

m-mueller678:
Another possible issue is recreating the future to lock the mutex on each poll call. This is not the intended way to use the mutex. The tokio documentation states that the future will usually not yield to the runtime if the access is uncontended, but that is explicitly not guaranteed. From a quick look at the Mutex code I could not find any situation where it would consistently yield on the first call. But if such a state is possible, this implementation would be wrong.

wngr:
> [..] actually reach into the host to read/write from the socket. [..]
>
> [..] This is a much larger effort, but I think its where we need to go for Tcp and Udp in turmoil.

While I'm not 100% clear which design you have in mind, `UdpSocket::readable` currently requiring a mutable reference makes turmoil's `{Tcp,Udp}Socket` implementations unusable for certain application designs. Whichever underlying implementation will be chosen, it would be really important to model the existing APIs from `tokio::net` 1-to-1.

Sorry for the drive-by comment.

mcches:
> > [..] actually reach into the host to read/write from the socket. [..]
> > [..] This is a much larger effort, but I think its where we need to go for Tcp and Udp in turmoil.
> 
> While I'm not 100% clear which design you have in mind, `UdpSocket::readable` currently requiring a mutable reference makes turmoil's `{Tcp,Udp}Socket` implementations unusable for certain application designs. Whichever underlying implementation will be chosen, it would be really important to model the existing APIs from `tokio::net` 1-to-1.
> 
> Sorry for the drive-by comment.

Thanks for stopping by! Whatever we do here will always ensure the API signatures mirror tokio 1:1.

src/net/udp.rs
```diff
@@ -247,6 +267,21 @@ impl UdpSocket {
         Ok((limit, origin))
     }
 
+    /// Tries to receive a single datagram message on the socket. On success,
+    /// appends to `buf` and returns the origin.
+    ///
+    /// If a message is too long to fit in the unfilled part of `buf` ,
+    /// excess bytes may be discarded.
```
rcoh, line 274: this also doesn't match the upstream — I think its critical to maintain precise fidelity. We'll need to keep the data around somewhere to avoid discarding it.

The only hack I can think of offhand is to keep a `Vec<u8>` or similar inside of of `Rx`. If there is overflow, write data there and immediately schedule a wakeup.

Ultimately, I think dropping data here is too much of a footgun to merge as is.

m-mueller678: Oh, I was under the impression that is what tokio does. Does it just deliver the remaining bytes on the next receive call then? I was sure I had had issues with truncated messages in tokio in the past, but it might not have been UDP. I agree, we should match tokio behavior exactly. There is already a buffer in there, so keeping it for the next call would be straight forward.

rcoh: ah! I read a lot of code to get to the bottom of this. This does in fact seem to be the (undocumented from Tokio's side) behavior. The exact semantics are platform dependent and also depend on the type of socket you are connected to. Windows will definitely drop them, it seems like Linux can do either.

I think Turmoil dropping these bytes _is_ the harshest and correct behavior (but maybe we should log as we do it?)

https://linux.die.net/man/2/recvfrom
> If a message is too long to fit in the supplied buffer, excess bytes may be discarded depending on the type of socket the message is received from.

rcoh: I think the right action here is:
- Capture that this is the behavior in a test
- Log or otherwise communicate to the turmoil user that this happened so they can get to the bottom of the problem more easily

Issue #212, opened 2025-01-29
Simulation shuts down without clients


m-mueller678:
I would expect it to print both lines, but I only get "before sleep".

```rust
use std::time::Duration;
use tokio::time::sleep;

fn main() {
    let mut sim = turmoil::Builder::new()
        .simulation_duration(Duration::from_secs(10))
        .build();
    sim.host("sleeper", move || async move {
        println!("before sleep");
        sleep(Duration::from_secs(1)).await;
        println!(" after sleep");
        Ok(())
    });
    sim.run().unwrap();
}
```

After some digging through the simulation code, it looks like a simulation without clients shuts down even if hosts are still running. If this behavior is intentional, I believe it should at least be more clearly communicated in the documentation. I am testing a consensus algorithm implementation, and some tests just have servers talking to each other with no clients involved.

mcches:
This is the desired behavior. I agree that it could be documented better. How about some additional doc on `sim.run()` and a `tracing` event if no clients are registered?

mcches:
I merged #213 , and we could consider a run() alternative that doesn't stop for client termination and instead runs for the sim duration.

mcches:
I'm closing for now, but feel free to open a more specific issue if you think we need an explicit API for running the sim without any clients. This would be equivalent to adding a client host that just sleeps for the desired duration.

PR #213, opened 2025-01-29
Early exit and tracing for no clients

mcches:
Improvemtns for #212

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -16,7 +16,7 @@ use super::Result;
 type Software<'a> = Box<dyn Fn() -> Pin<Box<dyn Future<Output = Result>>> + 'a>;
 
 /// Runtime kinds.
-enum Kind<'a> {
+pub enum Kind<'a> {
     /// A runtime for executing test code.
     Client,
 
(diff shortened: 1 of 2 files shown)


src/rt.rs
```diff
@@ -16,7 +16,7 @@ use super::Result;
 type Software<'a> = Box<dyn Fn() -> Pin<Box<dyn Future<Output = Result>>> + 'a>;
 
 /// Runtime kinds.
-enum Kind<'a> {
+pub enum Kind<'a> {
```
rcoh, line 19: Did you intentionally make these types public?

src/sim.rs
```diff
@@ -333,11 +335,21 @@ impl<'a> Sim<'a> {
         f(top.iter_mut())
     }
 
-    /// Run the simulation to completion.
+    /// Run the simulation until all client hosts have completed.
```
rcoh, line 338: Is there a standard way to keep the simulation alive? A client that just spins?

Could add a follow up comment along the lines of how to keep the sim running if it makes sense.

rcoh (approved):

src/rt.rs
mcches: Yes, the mod isn't pub so its just visible within the crate. I needed the kind to check for any client rts.

src/sim.rs
mcches: I think a sim with no clients is pretty niche. You'd essentially be running your sim and checking for something to err or panic. This was originally designed to have one or more hosts and then a client that drives the sim, or tests it. Registering a client that just polls forever would work, but I wonder if it'd be better to provide a different execution method instead, as an alternative to run().

PR #215, opened 2025-01-30
Update to rand 0.9

tottoto:
Updates to rand 0.9.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,8 @@ members = ["examples/*"]
 [dependencies]
 bytes = "1.4"
 indexmap = "2"
-rand = { version = "0.8.5", features = ["small_rng"] }
-rand_distr = "0.4.3"
+rand = { version = "0.9", features = ["small_rng"] }
+rand_distr = "0.5"
 regex = { version = "1", optional = true }
 scoped-tls = "1.0.1"
 tokio = { version = "1.25.0", features = ["full", "test-util"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -22,8 +22,8 @@ members = ["examples/*"]
 [dependencies]
 bytes = "1.4"
 indexmap = "2"
-rand = { version = "0.8.5", features = ["small_rng"] }
-rand_distr = "0.4.3"
+rand = { version = "0.9", features = ["small_rng"] }
```
mcches, line 25: I'm a little fuzzy on cargo resolver. Will this break consumers who also declare a dependency on 0.8?

tottoto: This seems a breaking change as `rand` updated `rand_core` and `rand_core::RngCore` is exposed at the `Builder::build_with_rng`.

https://docs.rs/turmoil/0.6.5/turmoil/struct.Builder.html#method.build_with_rng

mcches: Let's stash this one and we can bundle it when we go to 0.7?

teskje: Since we're going to have a breaking change anyway, can we consider removing `rand` from the public API? Instead of taking an rng value, could the builder take a seed instead? [That's what tokio does as well.](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.rng_seed) That would allow us to freely upgrade `rand` going forward, and it would also remove the need for the separate `build_with_rng` method.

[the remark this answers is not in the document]
tottoto: Updated.

Issue #207, opened 2025-01-02
Imprecise elapsed simulation time


nickgian:
`sim_elapsed` provides the simulation time elapsed as perceived by a host without a skew, by offsetting the host's time by the time the simulation started. However, when a host crashes time "freezes" for it and it remains skewed forever.
My title might be inaccurate - I am not sure if this is the desired behavior (to test impact of clock skew), but having a way for hosts to keep track of the simulation time elapsed (a global clock) simplifies testing global properties and debugging.

The example below demonstrates this:
```
   #[test]
    fn elapsed_time_across_crashes() -> Result {
        let tick_ms = 5;
        let mut sim = Builder::new()
            .tick_duration(Duration::from_millis(tick_ms))
            .build();


        sim.host("host", move || {
            async move {
                loop {
                    tokio::time::sleep(Duration::from_millis(1)).await;
                    println!("Time elapsed according to host: {}", sim_elapsed().unwrap().as_millis());
                }
            }
        });

        sim.host("host2", move || {
            async move {
                loop {
                    tokio::time::sleep(Duration::from_millis(1)).await;
                    println!("Time elapsed according to host2: {}", sim_elapsed().unwrap().as_millis());
                }
            }
        });

        println!("First run");
        sim.run()?;
        sim.crash("host");
        println!("Crashed host 1");
        sim.run()?;
        sim.bounce("host");
        println!("Bounced host 1");
        sim.run()?;

        Ok(())
    }
```

will output

```
First run
Time elapsed according to host: 1
Time elapsed according to host: 2
Time elapsed according to host: 3
Time elapsed according to host: 4
Time elapsed according to host2: 1
Time elapsed according to host2: 2
Time elapsed according to host2: 3
Time elapsed according to host2: 4
Crashed host 1
Time elapsed according to host2: 5
Time elapsed according to host2: 6
Time elapsed according to host2: 7
Time elapsed according to host2: 8
Time elapsed according to host2: 9
Bounced host 1
Time elapsed according to host: 6
Time elapsed according to host: 7
Time elapsed according to host: 8
Time elapsed according to host: 9
Time elapsed according to host2: 10
Time elapsed according to host2: 11
Time elapsed according to host2: 12
Time elapsed according to host2: 13
Time elapsed according to host2: 14
```

PR #216, opened 2025-02-02
Keep clocks of crashed hosts in-sync with the simulation.

nickgian:
This PR ticks the clock on hosts that are not actively running, so if they are ever restarted their clock will be synchronized with the time passed in the simulation. This allows us to use `sim_elapsed()` as a form of global clock, which every host can use. 

Relevant issue: #207

### DIFF
--- a/src/sim.rs
+++ b/src/sim.rs
@@ -383,11 +383,10 @@ impl<'a> Sim<'a> {
         // completes, extract the result and return early if an error is
         // encountered.
 
-        let mut running: Vec<_> = self
+        let (mut running, stopped): (Vec<_>, Vec<_>) = self
             .rts
             .iter_mut()
-            .filter(|(_, rt)| rt.is_software_running())
-            .collect();
+            .partition(|(_, rt)| rt.is_software_running());
         if self.config.random_node_order {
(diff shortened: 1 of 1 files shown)


src/sim.rs
```diff
@@ -906,6 +912,79 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn elapsed_time_across_crashes() -> Result {
```
marcbowes, line 916: This is a really great test, but I'd personally be inclined to keep it simpler (at the expense of coverage). I don't think showing host2 is unaffected by the crash is that helpful, nor do I think showing clock_1 falls behind is that helpful. Really, this could just be one host and the assertion on L974.

marcbowes: I understand you probably wrote the rest of it for your own learning/exploration of current behavior, but part of my inclination is based on what happens over time if you don't "spring clean" the tests. We land up with a bunch of ~100 LOC tests that really could be 20 LOC, which makes everything that much harder to grok.

This is not a review blocker, as we can always clean it later.

marcbowes (approved):

mcches (approved):

PR #218, opened 2025-02-08
Improve host ports exhausted panic message

quinnwerks:
I've added the hostname to the error message. While debugging an issue I found this useful. I want to put this back so others can benefit from this too.


quinnwerks:
@mcches when you have a chance. Thanks!

mcches (approved):

PR #224, opened 2025-02-23
TcpListener: drain all pending accepts

teskje:
This PR makes `TcpListener::accept` try to accept the next connection without waiting for a notify, if the previous connection couldn't be established because the peer already went away.

The change is meant to fix a bug where a pending connection sometimes wouldn't be accepted. Consider this series of events:

 * Server runs `TcpListener::accept`. There are no connections pending, so it starts waiting on the `notify`.
 * Client 1 sends a SYN, which gets pushed to `ServerSocket::deque` and `ServerSocket::notify` is notified.
 * Client 2 sends a SYN, which gets pushed to `ServerSocket::deque` and `ServerSocket::notify` is notified.
 * Client 1 goes away.
 * Server wakes up because the `notify` was notified.
 * Server tries to accept the connection from client 1, which fails because client 1 went away.
 * Server starts waiting on the `notify` again.

We end up in a situation where the connection from client 2 is still pending but the server doesn't accept it because it's waiting for a notification that was already sent.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -64,42 +64,48 @@ impl TcpListener {
     /// established, the corresponding [`TcpStream`] and the remote peer’s
     /// address will be returned.
     pub async fn accept(&self) -> Result<(TcpStream, SocketAddr)> {
-        loop {
+        let origin = loop {
             let maybe_accept = World::current(|world| {
                 let host = world.current_host_mut();
-                let (syn, origin) = host.tcp.accept(self.local_addr)?;
-
-                tracing::trace!(target: TRACING_TARGET, src = ?origin, dst = ?self.local_addr, protocol = %"TCP SYN", "Recv");
-
-                // Send SYN-ACK -> origin. If Ok we proceed (acts as the ACK),
-                // else we return early to avoid host mutations.
-                let ack = syn.ack.send(());
-                tracing::trace!(target: TRACING_TARGET, src = ?self.local_addr, dst = ?origin, protocol = %"TCP SYN-ACK", "Send");
-
-                if ack.is_err() {
-                    return None;
-                }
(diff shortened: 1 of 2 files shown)


teskje:
For sure, added one now!

src/net/tcp/listener.rs
```diff
@@ -64,10 +64,15 @@ impl TcpListener {
...
+            Now,
+            Wait,
+        }
+
         loop {
-            let maybe_accept = World::current(|world| {
+            let accept_result = World::current(|world| {
                 let host = world.current_host_mut();
-                let (syn, origin) = host.tcp.accept(self.local_addr)?;
+                let (syn, origin) = host.tcp.accept(self.local_addr).ok_or(Retry::Wait)?;
```
mcches, line 75: Could we return from the `World::current` closure here and avoid needing the enum? accept -> None signals the wait, syn.ack.send() -> err signals the retry, and the Ok path remains the same.

mcches (commented):
Looks pretty good! Let me know what you think about the accept simplification.

src/net/tcp/listener.rs
teskje: Just to check my understanding: You are suggesting that we shorten the `World::current` closure to already end here and then do the `syn.ack.send()` and testing of the result outside of it, where we can `continue` the loop if necessary?

That would work, but in the `Ok` case we're have to enter `World::current` again to get access to the `host` back, unless I'm missing something. Still sounds a bit simpler, I'll give it a try!

teskje: Pushed a change that gets rid of the enum and breaks out of the loop once we know we don't need to retry anymore. Lmk what you think!

mcches (approved):

PR #234, opened 2025-03-31
chore: add tcp server crash test

LucioFranco:


LucioFranco:
Yeah, I think this was just a test I was trying to see if there was a bug in turmoil or in my use case. It turns out that it was on my end so this can be closed.

PR #249, opened 2025-11-04
Add a `crate::since_epoch()` API to return logical time elapsed since epoch

raskyamazon:
This simplifies obtaining logical epoch time within a sim, which is useful for maintaining determinism in testing which relies on epoch time, e.g. for timestamps.


mcches (approved):
Thank you for the contribution!

PR #256, opened 2025-12-16
:sparkles: Add drop filter

hhalex:
Hello,

I added a configurable drop filter in Sim/Topology that runs before messages enter the network for any protocol/direction, plus helpers to set/clear it and a UDP test covering selective drops.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -460,6 +460,14 @@ impl Tcp {
         Some(sock.assign_seq())
     }
 
+    pub(crate) fn rollback_send_seq(&mut self, pair: SocketPair, seq: u64) {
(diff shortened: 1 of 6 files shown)


mcches (commented):
This seems like a nice change, but I'm curious if the existing mechanism could be modified instead?

https://docs.rs/turmoil/latest/turmoil/struct.SentRef.html

This could be updated to mark for drop.

mcches:
Sorry for the review delay. I was out for a bit. What's the state of this PR? I've got a couple of qs:

- Why wouldn't drop on SentRef work for this?
- How can we drop a tcp segment and the stream still works? I assume the seqno would be off.

PR #260, opened 2025-12-25
Let TCP peek methods take `&self`

teskje:
This changes the `TcpStream::peek` and the various other peek methods to take `&self` instead of `&mut self`, ensuring tokio API compatibility and making them usable in more contexts.

To achieve this, we wrap the `ReadHalf` internals in `Mutex` and `AtomicBool`, respectively. Note that we can't use `RefCell`/`Cell` here because we also require `ReadHalf` to be `Sync`.


mcches:
I'm currently working on an overhaul of net that fills all the API holes and adds more fidelity. Can you use a patch for now and then switch back to upstream once I get that done?

teskje:
For sure! Feel free to close this PR or leave it open for others to find the discussion, whatever suits you best.

PR #267, opened 2026-04-09
Drop parking_lot from tokio feature list

jgrund:
Replaces `tokio/full` with its enumerated contents minus `parking_lot`. Eliminates a wall-clock-dependent fair-scheduling path in turmoil's internal execution model, closing a non-determinism hole that affects any test with contention on internal tokio mutexes (Clock::Inner, blocking pool state, etc.).


mcches:
This seems fine, but with feature unification you'd have to ensure you weren't enabling `parking_lot` where ever you use turmoil.

mcches (approved):

PR #269, opened 2026-04-24
Send RST when a TcpStream is dropped with unread data

### LINKED ISSUE #158: Fix socket close behavior when there is unread data
[Placeholder]

mcches:
## Summary
- RFC 9293 §3.10.4: a receiver closing with unread bytes must send RST, not FIN. Previously the ReadHalf drop path fell through to the sibling WriteHalf's FIN, so peers saw a graceful EOF even when data was discarded.
- `ReadHalf::drop` now checks its partial buffer, the mpsc channel, and the host-side reorder buffer for undelivered `Data`. If any is found it emits `Segment::Rst` and tears down the socket locally, which also suppresses the sibling `WriteHalf`'s FIN (via `seq()` returning `Err`).
- Queued FINs are intentionally not treated as a reset condition — a graceful close followed by a drop stays graceful.

Fixes #158.

### DIFF
--- a/tests/net/tcp.rs
+++ b/tests/net/tcp.rs
@@ -1750,11 +1750,13 @@ fn close_with_unread_data_sends_rst() -> Result {
     });
 
     sim.client("client", async move {
-        let s = TcpStream::connect(("server", PORT)).await?;
+        let mut s = TcpStream::connect(("server", PORT)).await?;
 
-        // Give the server's data time to arrive in our receive queue before
-        // we drop without reading.
-        tokio::time::sleep(Duration::from_secs(1)).await;
+        // Block until at least one byte has arrived in our receive queue,
+        // then drop without consuming it. Peek stashes the bytes in the rx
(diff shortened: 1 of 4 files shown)


tests/net/tcp.rs
```diff
@@ -1717,3 +1717,49 @@ fn tcp_backpressure_on_loopback() {
...
+
+        Ok(())
+    });
+
+    sim.client("client", async move {
+        let s = TcpStream::connect(("server", PORT)).await?;
+
+        // Give the server's data time to arrive in our receive queue before
+        // we drop without reading.
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
atreyas: Could this use `has_buffered_data`?

mcches: this is AI slop. ill clean up and remove sleep

atreyas (approved):
Looks good to me.

PR #273, opened 2026-04-25
Convert to workspace and scaffold turmoil-net

mcches:
## Summary

- Move the existing crate under `crates/turmoil/` (history preserved via `git mv`) and promote the top-level `Cargo.toml` to a workspace manifest (`crates/*` + `examples/*`).
- Add an empty `crates/turmoil-net/` with the module skeleton for the simulated socket rewrite: `kernel/{socket,table,options,tcp,udp,uds}` and `shim/tokio/net/`, plus a `NetRuntime` trait sketch in `lib.rs` as the integration boundary with the host runtime.
- Example path deps rewritten to `../../crates/turmoil`. No behavior changes in `turmoil`.

The goal of the new crate is to mirror kernel socket APIs (AF_INET/AF_INET6/AF_UNIX, SOCK_STREAM/SOCK_DGRAM) as the authoritative implementation, with a thin tokio shim layered on top — same structural pattern as `fs/shim/`. `std::net` is intentionally not shimmed. Follow-ups will iterate on the kernel socket layer (TCP first, then UDP, then UDS).


.github/workflows/rust.yml
```diff
@@ -40,17 +40,17 @@ jobs:
       matrix :
         os: [ubuntu-latest]
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v5
       - name: Rust Toolchain
         uses: dtolnay/rust-toolchain@stable
       - name: Run cargo clippy
-        run: cargo clippy --all-targets -- --deny warnings
+        run: cargo clippy -p turmoil -p turmoil-net --all-targets -- --deny warnings
```
marcbowes, line 47: Why not everything?

PR #277, opened 2026-04-30
Add turmoil-net: deterministic simulated socket stack

mcches:
A POSIX-shaped simulated kernel + tokio-compatible shim, so tests can
swap `tokio::net` for `turmoil_net::shim::tokio::net` and run the
same code against a network that's fully deterministic and
controllable.

Scope:
- Tokio shim: TcpListener, TcpStream, UdpSocket matching tokio's
  public API. TcpStream supports split/into_split with ReadHalf/
  WriteHalf/OwnedReadHalf/OwnedWriteHalf. A parity check test
  (RUSTFLAGS=--cfg parity_tokio) asserts the trait surface against
  real tokio so drift is caught at build time.
- Kernel: per-host socket table; TCP handshake, streaming data,
  graceful/abortive close (FIN/RST), 4-tuple demux, backpressure,
  MSS-bounded segmentation. UDP send/recv/peek with connected-peer
  filtering. Cancellation-safe connect (FdGuard) and listener close
  that RSTs unaccepted children.
- Fabric: multi-host routing with a caller-driven sim clock;
  deterministic iteration via IndexMap throughout.
- Rules: packet-level fault injection with Pass/Deliver(delay)/Drop
  verdicts. Install permanently on Net or via RAII guard from
  EnterGuard / a free fn. Latency ships as the first concrete Rule.
- Fixtures: `fixture::lo` for single-host tests, `fixture::ClientServer`
  for N-server/one-client topologies. Both drive a paused current_thread
  runtime so tests complete deterministically with no wall-clock waits.
- Netstat: Linux-style socket snapshot (`netstat("host")`) for
  inspecting connection state from tests.
- DNS: hostnames auto-allocated to 192.168.x.x; literal IPs pass
  through. Used by Net::add_host and the shim's ToSocketAddrs.

67 tests covering TCP handshake/streaming/close, UDP roundtrip,
fault-injection rules, cancellation, multi-host delivery, and the
parity surface. README explains the motivation and rule model;
docs.rs landing points at the module map.

Higher-level fault primitives (named partitions, reorder, loss
distributions) and TCP retransmit will land in follow-ups — the rule
API and fabric clock are already in place to support them.


mcches:
This will be usable standalone and will soon replace the socket stack in turmoil.

PR #280, opened 2026-05-02
feat(fs): add short_read_probability fault injection

### LINKED ISSUE #279: unstable-fs: short-read fault injection for positional reads
### Use case

We're building a storage engine on top of turmoil's simulated filesystem
(`unstable-fs`, via `shim::tokio::fs`). Our `FileIo::read_at` contract is
"the returned buffer is exactly `len` bytes long, zero-padded past EOF" —
so every caller treats the result as a full buffer and `copy_from_slice`s
it into aligned destinations.

Under `O_DIRECT` on a real kernel, a `pread` on a regular file with
aligned args returns either the full count or EOF; transient shorts are
effectively impossible. But the sync backend also has a plain-`pread`
path (no `O_DIRECT`) where `EINTR` or similar shorts *are* possible, and
we've just landed a retry loop for that. We'd like turmoil to exercise it
deterministically.

### What we're asking for

A fault-injection knob (probability or predicate, matching the style of
#192) that can cause positional reads through turmoil's simulated fs —
`File::read`, `File::read_at`, and async equivalents — to return a
short count, with the remaining bytes left untouched or zeroed at the
sim's discretion. Callers that correctly handle shorts keep working;
callers that assumed full-read semantics (like us before the fix) fail
visibly under sim instead of only in prod.

Happy to send a PR if the shape of the API is agreed.

marcbowes:
## Summary

Adds a short-read fault-injection knob for the simulated filesystem, per #279.

`FsConfig::short_read_probability` causes positional reads through `File::read`, `File::read_at`, and their async equivalents to randomly return fewer bytes than requested with the configured probability. The tail of the buffer past the returned count is zeroed, so callers that incorrectly treat the result as a full buffer observe garbage instead of silently-correct data. Well-behaved callers that loop until the requested length is reached (or hit `Ok(0)` / EOF) keep working.

Injection fires on O_DIRECT reads too — signals (`EINTR`) can interrupt any blocking syscall, including aligned O_DIRECT `pread`.

```rust
let mut builder = Builder::new();
builder.fs().short_read_probability(0.1);  // 10% chance of a short read
```

Closes #279.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p turmoil -p turmoil-net -p turmoil-fs --all-targets -- --deny warnings`
- [x] `cargo test --workspace`
- [x] `cargo test --workspace --features regex`
- [x] `cargo test --workspace --features unstable-fs` (new tests: `short_read_100_percent`, `short_read_also_fires_for_direct_io`, `short_read_probability_statistical` at 0%/50%/100%)

### DIFF
--- a/crates/turmoil/src/fs/mod.rs
+++ b/crates/turmoil/src/fs/mod.rs
@@ -582,9 +582,9 @@ impl FsConfig {
     ///
     /// When set, reads through `File::read`, `File::read_at`, and their async
     /// equivalents may randomly return fewer bytes than requested, even when
-    /// more data is available. This simulates the short-read behavior of plain
-    /// `pread` on real kernels (e.g., interrupted syscalls, partial reads from
-    /// non-`O_DIRECT` file descriptors).
+    /// more data is available. This simulates the short-read behavior of
+    /// `pread` on real kernels (e.g., `EINTR` interrupting a blocking syscall).
+    /// O_DIRECT reads are not exempt — signals can short-circuit them too.
     ///
     /// The tail of the buffer past the returned count is zeroed, so callers
     /// that incorrectly treat the result as a full buffer observe garbage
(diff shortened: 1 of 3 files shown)


mcches (approved):
Nice!

PR #281, opened 2026-05-04
refactor(turmoil-net): let callers own the egress buffer

mcches:
`Kernel::egress`, `Fabric::egress_all`, and `EnterGuard::egress_all`
now append into a caller-supplied `&mut Vec<Packet>` instead of
returning a fresh `Vec` each call. The fixture `Scheduler` holds a
preallocated buffer and reuses it across ticks.

### DIFF
--- a/crates/turmoil-net/src/fabric.rs
+++ b/crates/turmoil-net/src/fabric.rs
@@ -105,9 +105,7 @@ impl Fabric {
     /// source host's `deliver` inline (inside `Kernel::egress`) and
     /// never appear in `out`.
     pub fn egress_all(&mut self, out: &mut Vec<Packet>) {
-        let ids: Vec<_> = self.hosts.keys().copied().collect();
-        for id in ids {
-            let host = self.hosts.get_mut(&id).expect("id from iteration");
+        for host in self.hosts.values_mut() {
             host.kernel.egress(out);
         }
     }
--- a/crates/turmoil-net/src/fabric.rs
+++ b/crates/turmoil-net/src/fabric.rs
@@ -98,19 +98,18 @@ impl Fabric {
         self.ip_to_host.get(&ip).copied()
     }
 
-    /// Drain outbound packets from every host, in host-insertion order.
-    /// The caller decides what happens to each packet — deliver, drop,
-    /// schedule for later. Loopback packets fold back through the
+    /// Drain outbound packets from every host, in host-insertion order,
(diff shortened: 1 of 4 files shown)


crates/turmoil-net/src/fabric.rs
```diff
@@ -98,19 +98,18 @@ impl Fabric {
...
+    /// appending to `out`. The caller decides what happens to each
+    /// packet — deliver, drop, schedule for later — and reuses the
+    /// buffer across ticks. Loopback packets fold back through the
     /// source host's `deliver` inline (inside `Kernel::egress`) and
-    /// never appear here.
-    pub fn egress_all(&mut self) -> Vec<Packet> {
-        let mut out = Vec::new();
+    /// never appear in `out`.
+    pub fn egress_all(&mut self, out: &mut Vec<Packet>) {
         let ids: Vec<_> = self.hosts.keys().copied().collect();
```
camshaft: I don't understand why we need to get a list of IDs here, only to iterate again for the lookup. Wouldn't it just be
```rust
for host in self.hosts.values_mut() {
```

crates/turmoil-net/src/fixture/scheduler.rs
```diff
@@ -57,14 +63,20 @@ impl Scheduler {
...
         }
 
-        for pkt in guard.egress_all() {
+        // Move the buffer out to a local: `schedule` (below) needs
+        // `&mut self`, which collides with a borrow of `self.egress`.
+        // Swapping it back at the end preserves the allocation.
+        let mut egress = std::mem::take(&mut self.egress);
+        guard.egress_all(&mut egress);
+        for pkt in egress.drain(..) {
             match guard.evaluate(&pkt) {
```
camshaft, line 72: At some point it would be nice to emit events or push the dropped packets to another queue so it's easier to debug when things go wrong. Not blocking here, though.

camshaft (approved):

crates/turmoil-net/src/fixture/scheduler.rs
mcches: This is just the simple scheduler used in the "batteries included" fixtures. For more complex harnesses that's totally doable now since they'll own this loop.

crates/turmoil-net/src/fabric.rs
mcches: fixed

PR #282, opened 2026-05-15
feat(turmoil): add simulated io_uring behind unstable-io_uring

marcbowes:
## Summary

- Adds `turmoil::io_uring`, a Rust-API-shape mirror of the [`io-uring = "0.7"`](https://docs.rs/io-uring/0.7) crate, gated by the new `unstable-io_uring` feature (depends on `unstable-fs`).
- Lets consumers swap their `use io_uring::*` imports behind a feature flag and run their full test suite under turmoil on macOS/Windows. The simulation samples completion latency from the existing fs distribution, RNG-shuffles CQE order, and shares state with `Fs` so ring writes participate in the same `pending`/`sync`/`crash` model as shim writes.
- Three-layer validation mirroring `turmoil-net`'s pattern: sim-only behavior tests, runtime conformance against the real Linux kernel io_uring, and `--cfg parity_io_uring` compile-time signature parity.

## Scope

Implemented:
- `IoUring`, `Builder` (`setup_sqpoll`/`setup_iopoll` rejected with `EINVAL`).
- SQ/CQ types: `submission()`, `submission_shared()`, `completion()`, `completion_shared()`. `CompletionQueue::next()` requires `sync()` first, matching real Linux semantics.
- Opcodes: `Read`, `Write`, `Fsync`, `AsyncCancel`. Anything else is unreachable through the builders.
- `Submitter::{submit, submit_and_wait, submit_with_args}`. `submit_with_args` validates `timespec` shape but does not bound waits (sync API; see doc).
- `AsyncFd<T: AsRawFd>` mirroring `tokio::io::unix::AsyncFd`, driven by per-ring `Notify` + `tokio::time::sleep_until` race in simulated time.
- `as_raw_fd()` on shim std and tokio `File`s, allocating from a `SIM_FD_BASE = 1<<30` range to avoid colliding with real fds.

Out of scope (rejected with `EINVAL` at submit, or simply not exposed):
- SQPOLL, IOPOLL, registered files/buffers, multishot, linked SQEs, BufRing, ReadFixed/WriteFixed.

Behavioral divergences from real Linux that are *documented at the call site*:
- `File::drop` between submit and completion turns the in-flight op's CQE into `-EBADF` (real kernel keeps an `f_count` ref).
- Page cache: two reads at the same offset in one batch — second sees a hit because we insert at submit time. Matches the existing tokio shim model.
- `CompletionQueue::next()` runs the side effect inside the iterator (real kernel pre-materializes).

## Validation

| | macOS | Linux |
|---|---|---|
| `cargo test --features unstable-io_uring` | 271/271 ✓ | 276/276 ✓ |
| `cargo test --features unstable-io_uring,real-uring-conformance` | (real arm cfg-hidden) | 283/283 ✓ |
| `RUSTFLAGS=--cfg=parity_io_uring cargo check --tests` | (parity cfg-hidden) | ✓ |
| `cargo clippy --all-targets -- --deny warnings` | ✓ | ✓ |
| `cargo fmt --check` | ✓ | ✓ |

The conformance suite runs five test bodies (read, write+fsync, EBADF, ENOENT cancel, multi-CQE batch) against both the sim and the real Linux kernel io_uring. All five behaviors agree across backends.

The parity check caught one drift on first run (a `Builder` that derived `Debug` where the real crate doesn't) — fixed pre-merge.

## Test plan

- [x] Sim-only tests pass on macOS.
- [x] Sim-only tests pass on Linux.
- [x] Real-arm conformance compiles and runs against actual Linux kernel io_uring (10/10 conformance tests).
- [x] `RUSTFLAGS=--cfg=parity_io_uring` build resolves shim against real `io-uring` crate without errors.
- [x] All four feature configs build cleanly: `<none>`, `unstable-fs`, `unstable-io_uring`, `unstable-io_uring,real-uring-conformance`.
- [x] `cargo clippy -- --deny warnings` clean across all configs.
- [x] No regressions in existing turmoil-net or fs tests.

### DIFF
--- a/crates/turmoil/src/io_uring/cqueue.rs
+++ b/crates/turmoil/src/io_uring/cqueue.rs
@@ -1,0 +1,149 @@
+//! Completion queue types mirroring [`io_uring::cqueue`](https://docs.rs/io-uring/0.7/io_uring/cqueue).
+
+use crate::fs::FsContext;
+use std::os::fd::RawFd;
+
+/// Sealed trait selecting between [`Entry`] (default) and the larger
+/// `Entry32` shape exposed by the real crate. Only `Entry` is
+/// implemented in the simulation.
+pub trait EntryMarker: private::Sealed + Sized {}
+
+/// One completion queue entry. Yielded by iterating a
+/// [`CompletionQueue`].
+#[derive(Clone, Debug)]
+pub struct Entry {
+    pub(crate) user_data: u64,
+    pub(crate) result: i32,
+    pub(crate) flags: u32,
+}
+
+impl Entry {
+    /// User-data tag set on the originating SQE.
+    pub fn user_data(&self) -> u64 {
+        self.user_data
+    }
+
+    /// Op result: non-negative on success (e.g. bytes read), negative
+    /// `errno` on failure.
+    pub fn result(&self) -> i32 {
+        self.result
+    }
+
+    /// Per-CQE flags. The simulation always reports zero.
+    pub fn flags(&self) -> u32 {
+        self.flags
+    }
+}
+
+impl EntryMarker for Entry {}
+
+/// Completion queue handle. Iterating it yields any CQEs whose
+/// scheduled completion time has elapsed in simulated time, in
+/// RNG-shuffled order.
+///
+/// # Real-Linux fidelity: `sync()` is required before iterating
+///
+/// On real Linux, [`Self::sync`] copies the kernel's CQ head pointer
+/// into userspace; without it, `next()` returns `None` even when the
+/// kernel has posted CQEs. The simulation enforces the same contract:
+/// a freshly-constructed `CompletionQueue` has `visible == 0`, and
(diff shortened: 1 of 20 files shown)


crates/turmoil/src/io_uring/cqueue.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub(crate) fn new(ring_fd: RawFd) -> Self {
+        Self {
+            ring_fd,
+            _lifetime: std::marker::PhantomData,
+        }
+    }
+
+    /// Mirror the real crate's API. The simulation eagerly materializes
+    /// CQEs in `next()`, so this is a no-op.
+    pub fn sync(&mut self) {}
```
marcbowes: Will fix in the next revision

marcbowes:
discussed offline with @mcches 

This is net new code under an unstable feat flag. Merging so we can continue with the crate breakup.

PR #283, opened 2026-05-25
feat: extract turmoil-fs and turmoil-io_uring as standalone crates

marcbowes:
## Summary

- Lifts `unstable-fs` out of `turmoil` into a new publishable `turmoil-fs 0.1.0` crate.
- Lifts `unstable-io_uring` out of `turmoil` into a new publishable `turmoil-io_uring 0.1.0` crate, with an optional `fs` feature that pulls in `turmoil-fs`.
- Public API surface unchanged — `turmoil::fs::*`, `turmoil::io_uring::*`, `turmoil::FsConfig` etc. continue to resolve via re-export, so existing consumers see no source-level changes.

## How it's wired

Cross-crate state access uses installable host-accessor closures (the same shape `turmoil-net` uses for its `CURRENT` thread-local). The sister crates do not import `turmoil`. `turmoil`'s `Builder::build` installs the accessors once per process:

- `turmoil_fs::install_host_accessor` / `install_corruption_hook`
- `turmoil_io_uring::host::install_host_accessor` (io_uring-only borrow)
- `turmoil_io_uring::host::install_combined_accessor` (combined fs+io_uring borrow, used by `submit::schedule_pending` and `cqueue::CompletionQueue::next` — lock order: fs first, then io_uring)

A few `Fs` fields/methods are widened to \`pub\` so \`turmoil-io_uring\` can call them via \`&mut Fs\`. Documented as the published-but-unstable sister-crate surface; not stable for end users. The \`unstable-*\` features remain explicitly outside semver.

## Decoupling

- \`Fs\` no longer holds \`io_uring_rings\`. Ring storage moved to a new \`IoUringHostState\` owned by \`Host\` under the \`unstable-io_uring\` feature.
- \`Sim::crash\` now calls public per-subsystem crash APIs (\`Fs::crash\`, \`IoUringHostState::crash\`) instead of reaching into fs internals.
- The fs↔io_uring forward call (\`PendingApply::execute(&mut Fs)\`) is the only legitimate cross-crate contract and is preserved.

## \`turmoil-io_uring\` without \`fs\`

Without the \`fs\` feature, only the data-shape types (\`Entry\`, \`Flags\`, \`opcode::*\`, \`squeue::Entry\`, \`cqueue::Entry\`) compile — useful for parity tests against the real \`io-uring\` crate. The runtime (\`IoUring\`, \`Submitter\`, \`AsyncFd\`, \`SubmissionQueue\`, \`CompletionQueue\`) is gated behind \`fs\`.

## CI / release-plz

- \`release-plz.toml\`: drop \`release = false\` for \`turmoil-fs\`. Both new crates now participate in the release flow.
- First publish caveat: release-plz can't create new crates on crates.io, so \`cargo publish -p turmoil-fs\` and \`cargo publish -p turmoil-io_uring\` will need to be run manually once after this PR merges. Subsequent bumps are automatic.
- \`.github/workflows/rust.yml\`: new \`unstable-features\` job exercises \`unstable-fs\`, \`unstable-io_uring\` (sim + real-Linux ring), \`unstable-barriers\`, the \`parity_io_uring\` cfg check, and confirms \`turmoil-io_uring\` builds without the \`fs\` feature. Clippy job extended to include \`turmoil-io_uring\`.

## Forward-looking

When io_uring later wants to dispatch to net (real Linux io_uring submits SEND/RECV/CONNECT/ACCEPT against TCP sockets), two clean follow-ups remain open: (a) an \`IoBackend\` trait in \`turmoil-io_uring\` that fs/net both impl with fd-based dispatch; (b) push fd ownership into a \`turmoil\`-core fd registry. The current shape narrowed fs↔io_uring coupling to a single forward call, so either path is a focused refactor.

## Test plan

- [x] \`cargo fmt --check\`
- [x] \`cargo clippy -p turmoil -p turmoil-net -p turmoil-fs -p turmoil-io_uring --all-targets -- --deny warnings\`
- [x] \`cargo test --workspace --features regex\`
- [x] \`cargo test -p turmoil --features unstable-fs --test fs\` — 104 passed
- [x] \`cargo test -p turmoil --features unstable-io_uring --test fs\` — 132 passed
- [x] \`cargo test -p turmoil --features unstable-io_uring --test io_uring_conformance\` — 5 passed
- [x] \`cargo test -p turmoil --features unstable-io_uring,real-uring-conformance --test io_uring_conformance\` — 10 passed (real Linux ring)
- [x] \`cargo test -p turmoil --features unstable-barriers --test barriers\` — 11 passed
- [x] \`RUSTFLAGS=--cfg=parity_io_uring cargo check -p turmoil --features unstable-io_uring --tests\`
- [x] \`cargo build -p turmoil-io_uring\` (no \`fs\` feature — verifies the data-shape-only mode compiles)
- [x] \`cargo build --workspace\` (all examples still build)


mcches (changes requested):
This looks great, but I think we should move to a single layer of state for setting up the fs. We don't need that accessor stuff - there isn't anything global for this like what is required in net.

mcches (approved):
I'm going to merge this in and iterate before publishing.

PR #284, opened 2026-07-16
refactor(fs)!: simplify setup — Fs is now a handle, enter takes no context

mcches:
## Summary

- `Fs` is now a thin `Clone` handle wrapping `Arc<Mutex<FsState>>` — the mutex is hidden from callers.
- Setup is just `let fs = Fs::builder().seed(42).build(); let _guard = fs.enter();`
- Removed `EnterCtx` — simulated time and corruption hook live on `FsState` directly. Three thread-locals collapsed to one.
- `FsState` (formerly `Fs`) is the internal state struct, still pub for turmoil-io-uring's async completion path.

## Test plan

- [x] `cargo test --workspace` — all pass
- [x] `cargo test -p turmoil --features unstable-io_uring,unstable-barriers --test fs --test io_uring_conformance` — 137 tests pass
- [x] `cargo clippy` clean across all four crates
- [x] `cargo fmt --check` clean

### DIFF
--- a/crates/turmoil-fs/src/lib.rs
+++ b/crates/turmoil-fs/src/lib.rs
@@ -26,26 +26,32 @@
 //!   - Pending writes to synced files are lost; durable data survives
 //! - Each host has its own isolated filesystem namespace
 //!
-//! # Embedding
+//! # Setup
 //!
-//! `turmoil-fs` is normally consumed via `turmoil`, which wires it into
-//! the per-host simulation state. If you are building a custom harness,
-//! call [`install_host_accessor`] once per simulation to plug a callback
-//! that hands [`FsContext::current`] (and [`FsContext::current_if_set`])
-//! the current host's filesystem `Arc<Mutex<Fs>>`, simulated time, and
-//! RNG.
+//! ```ignore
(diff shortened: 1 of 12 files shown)


crates/turmoil-fs/src/lib.rs
```diff
@@ -229,15 +177,13 @@ struct WorkerContext {
 ///
 /// # Thread Safety
 ///
-/// The handle holds an `Arc<Mutex<Fs>>`, so it is safe to send to other threads.
+/// The handle clones the inner `Arc`, so it is safe to send to other threads.
 /// The mutex ensures exclusive access to the filesystem during operations.
 /// However, users should be aware that concurrent access from multiple threads
 /// may lead to non-deterministic behavior in tests.
 pub struct FsHandle {
```
Benjscho: Do we need both an FsHandle and an Fs if they have the same inner? I'm a bit confused about the distinction of when you'd want one vs the other

Benjscho (approved):

crates/turmoil-fs/src/lib.rs
mcches: I pushed a commit to simplify this further.

